PR #7, opened 2025-09-16
Add support for lyric files

potelux:
This draft PR introduces protocol support for lyric files.

There are several decisions to be made:
- Is support for various lyric file types necessary?  Or should just LRC be supported?
- What data should be sent to the client and when?
  - Should the server parse the file and send the timestamped lyrics?  
    - This would be great for low power devices or devices with limited screen space.
  - Should the entire LRC file be sent to the device to parse on its own? 
    - This would be great for a device that wants to display the previous lyrics and future lyrics alongside the current lyrics.  Like this:
484230538-0bd30346-5320-4f9d-a85f-b0be8948cd6b


README.md
```diff
@@ -359,6 +362,7 @@ The `metadata` object in [`session/update`](#server--client-sessionupdate) has t
   - `album_artist?`: string | null
   - `album?`: string | null
   - `artwork_url?`: string | null
+  - `lyric_url?`: string | null
```
marcelveldt, line 365: should the client receive the lyrics from an HTTP location or should we instead send the lyrics as binary data ?

potelux:
This can be explored again in the future

PR #20, opened 2025-10-07
Add `client_id` and `name` to the `_resonate` mDNS service type

maximmaxim345:


maximmaxim345:
I'm still not sure if this is required.
My main idea was that servers, that have a specific client disabled know what clients to skip.

But the server could just as well establish a websocket connection, receive the `client/hello` message, and immediately cancel the connection. Client initiated connections will already need to be handled this way, since they don't advertise the  `_resonate` mDNS type.

PR #22, opened 2025-10-07
Create new `artwork` role

maximmaxim345:
This PR introduces a dedicated `artwork` role as discussed on [discord](https://discord.com/channels/753947050995089438/1369735163629801532/1424502752310198284), separating artwork display from the `metadata` role.

The `metadata` role now only handles text information (title, artist, album, etc.), while `artwork` handles all image display.

Additionally, as discussed on [discord](https://discord.com/channels/753947050995089438/1369735163629801532/1421197479566643333):
- The `stream/request-format` message is extended to support both `player` and `artwork` roles, allowing artwork clients to dynamically adapt image formats and resolutions.
- We now assume that the server supports all image formats

Empty binary messages are now defined to clear artwork image.

This PR also implements another Idea, artwork channels:
- Artwork clients can support 1-4 independent channels, allowing simultaneous display of multiple images (e.g., album art on one channel, artist photos on another). Each channel can specify its source type (`album` or `artist`), format (`jpeg`, `png`, `bmp`), and resolution.
- This also allows client use cases such as showing a smaller thumbnail image while a full screen player is minimized. Without having to downscale the image themselves.
- Binary message types are updated: types 2-5 for artwork channels 0-3, type 6 for visualizer (previously type 3). A later PR will properly clean up binary message types by grouping them by role.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,10 +387,9 @@ This section describes messages specific to clients with the `artwork` role, whi
 The `artwork_support` object in [`client/hello`](#client--server-clienthello) has this structure:
 
 - `artwork_support`: object
-  - `support_picture_formats`: object[] - list of supported artwork formats in priority order (must have at least one format, first is preferred)
-    - `format`: 'jpeg' | 'png' | 'bmp' - image format identifier
-    - `media_width`: integer - max width in pixels
-    - `media_height`: integer - max height in pixels
+  - `format`: 'jpeg' | 'png' | 'bmp' - image format identifier
+  - `media_width`: integer - max width in pixels
+  - `media_height`: integer - max height in pixels
 
 **Note:** The server will scale images to fit within the specified dimensions while preserving aspect ratio.
 
--- a/README.md
+++ b/README.md
@@ -382,55 +382,71 @@ Clients can calculate the current track position at any time using the last rece
 ## Artwork messages
 This section describes messages specific to clients with the `artwork` role, which handle display of artwork images. Artwork clients receive images in their preferred format and resolution.
 
+**Channels:** Artwork clients can support 1-4 independent channels, allowing them to display multiple related images. For example, a device could display album artwork on one channel while simultaneously showing artist photos or background images on other channels. Each channel operates independently with its own format, resolution, and source type (album or artist artwork).
+
 ### Client → Server: `client/hello` artwork support object
 
 The `artwork_support` object in [`client/hello`](#client--server-clienthello) has this structure:
 
 - `artwork_support`: object
-  - `format`: 'jpeg' | 'png' | 'bmp' - image format identifier
-  - `media_width`: integer - max width in pixels
-  - `media_height`: integer - max height in pixels
+  - `channels`: object[] - list of supported artwork channels (length 1-4), array index is the channel number
+    - `source`: 'album' | 'artist' - artwork source type
+    - `format`: 'jpeg' | 'png' | 'bmp' - image format identifier
+    - `media_width`: integer - max width in pixels
+    - `media_height`: integer - max height in pixels
 
-**Note:** The server will scale images to fit within the specified dimensions while preserving aspect ratio.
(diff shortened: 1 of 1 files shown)


maximmaxim345:
In case the channel concept makes the `artwork` role too complicated, I can also revert the last commit that added that.

maximmaxim345:
Another idea for the channels concept: Applying costly filters like blur on the server to allow low powered clients to still have a fancy UI.

README.md
```diff
@@ -169,10 +170,11 @@ Players that can output audio should have the role `player`.
 - `supported_roles`: string[] - at least one of:
   - `player` - outputs audio
   - `controller` - controls the current Resonate group
-  - `metadata` - displays metadata
+  - `metadata` - displays text metadata
```
kahrendt, line 173: - `metadata` - displays text metadata describing the currently playing audio

README.md
```diff
@@ -346,30 +356,7 @@ Group state update.
...
-- `metadata`: object
-  - `art_format`: 'bmp' | 'jpeg' | 'png' - format of the encoded image
-
-### Server → Client: `stream/update` metadata object
-
-The `metadata` object in [`stream/update`](#server--client-streamupdate) has this structure with delta updates:
-
-- `metadata`: object
-  - `art_format`: 'bmp' | 'jpeg' | 'png' - format of the encoded image
+This section describes messages specific to clients with the `metadata` role, which handle display of track information and playback state. Metadata clients receive session updates with track details.
```
kahrendt, line 359: Is "playback progress" better than "playback state" here? Playback state could be read as describing whether its playing, paused, etc...

README.md
```diff
@@ -392,16 +379,75 @@ Clients can calculate the current track position at any time using the last rece
...
+    - `media_height`: integer - max height in pixels
+
+**Note:** The server will scale images to fit within the specified dimensions while preserving aspect ratio. Clients can support 1-4 independent artwork channels depending on their display capabilities. The channel number is determined by array position: `channels[0]` is channel 0 (binary message type 2), `channels[1]` is channel 1 (binary message type 3), etc.
+
+### Client → Server: `stream/request-format` artwork object
+
+The `artwork` object in [`stream/request-format`](#client--server-streamrequest-format) has this structure:
+
+- `artwork`: object
+  - `channels?`: object[] - configuration for artwork channels, array index is the channel number
```
kahrendt, line 405: Hypothetical here. The client sends 3 different channels in the original hello message, album, artist, and low resolution album. What happens if this request-format message has only one format listed? How would the server know which channel it should change the format to? We could require that the client send the same number of artwork channels as in the hello message, or we could add a specific channel field to this object to override that specific channel

kahrendt (changes requested):
I do like this channel implementation! We just need to make sure there isn't ambiguity in the spec around the request-format message.

README.md
```diff
@@ -393,16 +384,76 @@ Clients can calculate the current track position at any time using the last rece
...
+    - `media_height`: integer - max height in pixels
+
+**Note:** The server will scale images to fit within the specified dimensions while preserving aspect ratio. Clients can support 1-4 independent artwork channels depending on their display capabilities. The channel number is determined by array position: `channels[0]` is channel 0 (binary message type 2), `channels[1]` is channel 1 (binary message type 3), etc.
+
+### Client → Server: `stream/request-format` artwork object
+
+The `artwork` object in [`stream/request-format`](#client--server-streamrequest-format) has this structure:
+
+- `artwork`: object
+  - `channels?`: object[] - configuration for artwork channels
```
kahrendt, line 410: We should drop the ``channels`` list altogether to match the player's ``request-format`` message which only sends one setting. If the client wants to change the settings for multiple channels, it can send multiple request-format messages.

kahrendt (changes requested):

README.md
maximmaxim345: Done.

But I'm thinking about `request-format`, do we want to also add a `none` type for `source`.
Or some other way to allow a client to temporarily disable receive artwork in case its not required?

The use case I'm thinking of is a dynamic UI where the artist picture is only shown when clicking on it or something similar.
Then the client doesn't need to store the artist picture in RAM while it isn't required. Once the user clicks on the client request the `source` to be changed from `none` back to `artist`, so the server sends the artwork data. No need to always keep it loaded.

maximmaxim345: Done. Also added a note that `stream/request-format` will immediately resend the artwork binary data since that was not defined yet.

PR #35, opened 2025-11-17
Define how group volume is calculated and applied

maximmaxim345:
Define group volume as the average of all player volumes. When setting group volume, apply the delta to each player to preserve relative volume levels between players.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,8 @@ Control the group that's playing and switch groups. Only valid from clients with
   - `volume?`: integer - volume range 0-100, only set if `command` is `volume`
   - `mute?`: boolean - true to mute, false to unmute, only set if `command` is `mute`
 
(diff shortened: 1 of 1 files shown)


maximmaxim345:
There is a limitation with the algorithm described in this PR:
when individual players hit the 0% or 100% boundaries, the delta-based approach may not achieve the exact requested group volume.

For example:
- Player A is at 0%, Player B is at 100%
- The group volume is 50% (average)
- User sets group volume to 100%
- Delta = 100 - 50 = +50
- Apply delta to each player:
  - Player A: 0 + 50 = 50%
  - Player B: 100 + 50 = 150% → clamped to 100%
- The resulting group volume = (50 + 100) / 2 = 75% instead of the requested 100%

This happens because the delta gets "lost" when a player hits the volume ceiling or floor.
Is this an acceptable trade of, or should we compensate by redistributing the lost delta to players that haven't reached their boundaries?

kahrendt (approved):

PR #43, opened 2025-11-19
Add multi-server support through `client/goodbye` message

maximmaxim345:
Enables clients to switch between servers in multi-server environments.

Servers now indicate their connection intent via `connection_reason` in `server/hello`, allowing clients to decide which server to keep when a second one connects. The new `client/goodbye` message lets clients explain why they're disconnecting, helping servers respond appropriately.

For client-initiated connections, multi-server behavior is left implementation-defined since servers cannot reclaim clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,25 @@ The server discovers available clients through mDNS and connects to each client 
 
 **Note:** Do not manually connect to servers if you are advertising `_resonate._tcp`.
 
+#### Multiple Servers
+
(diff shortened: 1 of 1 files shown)


kahrendt (approved):
Looks good. This will take some time for me to implement properly in my ESPHome client. As a stopgap, I will add a way to just trigger a goodbye message for the ``another_server reason``, so we can just add template button in ESPHome to trigger that for development.

PR #67, opened 2026-01-30
Allow per-player delay configuration

maximmaxim345:
Players report `static_delay_ms` in `client/state` (always present, 0-5000ms range). Servers can adjust it via `set_static_delay` command for players that list it in `supported_commands` in the player roles `client/state` object.

Clients subtract `static_delay_ms` from server timestamps before scheduling playback (a 300ms delay means external equipment has 300ms latency, so playback starts earlier). Servers factor it into buffer calculations to maintain headroom. Value persists locally across reboots.

## Example implementations

| Scenario | `static_delay_ms` | `set_static_delay` supported |
|----------|-------------------|------------------------------|
| DIY speaker without pre-calibrated delay | User adjustable | Yes |
| Ready-made speaker with factory-measured delay | 0 | No |
| Active speakers connected to pre-calibrated audio port | Delay matching speaker latency | Yes |
| Active speakers with delay measured outside Sendspin | Read-only value for buffer optimization | No |

## Limitations

- No way to reduce server buffer headroom for devices with inherent internal delay (e.g., 100ms DAC/DSP). Since 0 must remain the default baseline, that 100ms cannot be reclaimed. We will accept that players with a high latency DAC/DSP will need to advertise a larger `buffer_capacity` and compensate that themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ Binary audio messages contain timestamps in the server's time domain indicating 
 - The server is unaware of individual client synchronization accuracy - it simply broadcasts timestamped audio
 - The server sends audio to late-joining clients with future timestamps only, allowing them to buffer and start playback in sync with existing clients
 - Audio chunks may arrive with timestamps in the past due to network delays or buffering; clients should drop these late chunks to maintain sync
+- Clients subtract their [`static_delay_ms`](#client--server-clientstate-player-object) from server timestamps before scheduling playback
+- Servers factor in each client's `static_delay_ms` when calculating how far ahead to send audio, keeping effective buffer headroom constant
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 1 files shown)


kahrendt (approved):
LGTM

README.md
```diff
@@ -464,6 +466,10 @@ State updates must be sent whenever any state changes, including when the volume
 - `player`: object
   - `volume?`: integer - range 0-100, must be included if 'volume' is in `supported_commands` from [`player@v1_support`](#client--server-clienthello-playerv1-support-object)
   - `muted?`: boolean - mute state, must be included if 'mute' is in `supported_commands` from [`player@v1_support`](#client--server-clienthello-playerv1-support-object)
+  - `static_delay_ms`: integer - static delay in milliseconds (0-5000), always required for players
+  - `supported_commands?`: string[] - subset of: 'set_static_delay'
```
balloob-travel, line 470: Since we are not final yet, why do we make this an opt-in thing, and not a default feature clients have to implement?

kahrendt: On a VPE, if they are using the internal speaker, we know the exact delay due to the internal DAC. If someone plugs in something into the aux port, we no longer know the exact delay. So the client will then say I now support setting the static delay (and will start using it if one is already saved). If the user unplugs the aux cord, then the client updates and says it is not longer possible to configure this (and no longer applies the saved value)

balloob-travel: Damn it, I am pretty sure you have told me this before. Sorry

kahrendt: I had to ask Maxim to remind me about it to before I merged it ;) I forgot it as well even though he explained it to me several times

PR #69, opened 2026-02-26
Add player timing capabilities

maximmaxim345:
There were 2 major gaps in the `player@v1` role this PR aims to solve:
- There is no minimum buffer size for live streams and other real-time content, potentially causing buffer underruns depending on the server implementation.
- Audio data could be sent immediately after starting a stream, causing the beginning of the audio to be cut off.

Both issues are resolved by adding `required_lead_time_ms` and `min_buffer_ms` to the `client/state` player object. The server is then responsible for ensuring that both constraints are respected.
Network latency can now also be factored into `required_lead_time_ms` and `min_buffer_ms` in real time, also allowing for high-latency clients (like a player on a mobile network) under changing network conditions.


README.md
```diff
@@ -451,10 +453,23 @@ The `player@v1_support` object in [`client/hello`](#client--server-clienthello)
     - `sample_rate`: integer - sample rate in Hz (e.g., 44100)
     - `bit_depth`: integer - bit depth for this format (e.g., 16, 24)
   - `buffer_capacity`: integer - max size in bytes of compressed audio messages in the buffer that are yet to be played
+  - `required_lead_time_ms`: integer - minimum startup lead time in milliseconds (e.g., codec init, decode warmup, audio backend buffering, DAC latency). Measure this from server transmit time of the start/restart trigger message ([`stream/start`](#server--client-streamstart) or [`stream/clear`](#server--client-streamclear)) to the timestamp of the first subsequent audio chunk.
+  - `min_buffer_ms`: integer - requested minimum ongoing buffer duration in milliseconds during playback (primarily for live streams), used to absorb network jitter and continuous-playback pipeline delays.
```
kahrendt: I'm a bit conflicted on these being statically set in the hello message (I know the PR mentions we make this assumption). In ESPHome, the very first run has at least double the lead time whereas subsequent lead times are much shorter (though this depends on the exact configuration, so it's tricky). For the min buffer, this could easily change on a cell connection at various points in time depending on signal strength. This one seems a lot harder for the server to dynamically update with though if we even had some mechanism to update it.

Should we consider allowing the hello message to be sent again in the middle of a live connection to update values (this has also come up for devices that want to switch between allowing volume control from the server or not; e.g., you plug a VPE into an amp and want the amp to handle the reset so you disable volume control on the server.

README.md
```diff
@@ -451,10 +453,23 @@ The `player@v1_support` object in [`client/hello`](#client--server-clienthello)
...
+
+**Server behavior:**
+- For startup/restart timing, compute per-player send-ahead using `required_lead_time_ms + static_delay_ms`.
+- For grouped startup/restart, use a common send-ahead of `max(required_lead_time_ms + static_delay_ms)` across grouped players, plus optional `sync_guard_ms`.
+- For ongoing playback timing, compute per-player send-ahead using `min_buffer_ms + static_delay_ms`.
+- For live streams or other real-time content with grouped playback, use a common ongoing send-ahead of `max(min_buffer_ms + static_delay_ms)` across grouped players, plus optional `sync_guard_ms`.
+- For live streams, keep each player's minimum buffer duration at or above `min_buffer_ms` when possible, capped by the maximum buffer size advertised in `buffer_capacity`.
+- For buffered streams, prefer filling each player's queue near `buffer_capacity` to maximize stability.
+- `buffer_capacity` is a hard per-player byte limit; servers should not send data that would cause a player's queued compressed audio to exceed this limit.
+
```
kahrendt, line 469: What happens in a live stream when a new player joins an already active group and it has a much higher `min_buffer_ms` value? What if a player leaves a group and now the max `min_buffer_ms` is much lower?

maximmaxim345: > What happens in a live stream when a new player joins an already active group and it has a much higher `min_buffer_ms` value?

For buffered streams it doesn't really matter.
But for live streams the server needs to pause the stream of a moment to build up the buffer. But I think the "when possible" was worded too loose. Changed it to "servers must maintain each player's ongoing buffer duration at or above its min_buffer_ms" to be more strict.

> What if a player leaves a group and now the max `min_buffer_ms` is much lower?

This is open to the server implementation. I guess most servers will just continue playing as before.
But servers that know that they need the lowest possible latency can decrease the buffer by clearing the buffer a bit to have a lower latency.

README.md
kahrendt: I've been thinking more about this. I can't come up with a way, at least in the ESPHome implementation, to actually determine these numbers. At best, I can make an educated case, but some of it will depend on the chip (ESP32 vs ESP32S3) and how many other things are running (microWakeWord/voice assistant stuff) and that makes it even harder to get good values that aren't overly conservative.

However, it is easy to determine this stuff empirically if we play audio once or have the connection open. I can just measure directly and add in a small margin. The client can also easily get an estimate for the latency from the time messages by just considering worst-case RTT (or something like 95th percentile).

So I don't know how to best do this in the hello message! I could save these details to the flash, but I don't know if I trust network latency to always be consistent day to day/across reboots. I don't know how a phone app could do it well either based on historical values... what if you are connected to local wifi vs on a cellular network

maximmaxim345: You have a point, I moved it from the hello to the state message.
For clients that set a static value it's still essentially the same, since sending a initial state update after connecting is already required by the spec.

This also removes the limitation I previously noted:
> Network latency is not measured and remains static throughout the lifetime of the connection.
While this is sufficient for almost all real-world scenarios, including mobile-network players, it is still a minor limitation.

README.md
```diff
@@ -469,10 +484,14 @@ State updates must be sent whenever any state changes, including when the volume
   - `volume?`: integer - range 0-100, must be included if 'volume' is in `supported_commands` from [`player@v1_support`](#client--server-clienthello-playerv1-support-object)
   - `muted?`: boolean - mute state, must be included if 'mute' is in `supported_commands` from [`player@v1_support`](#client--server-clienthello-playerv1-support-object)
   - `static_delay_ms`: integer - static delay in milliseconds (0-5000), always required for players
+  - `required_lead_time_ms`: integer - minimum startup lead time in milliseconds (e.g., codec init, decode warmup, audio backend buffering, DAC latency), always required for players. Measured from server transmit time of the start/restart trigger ([`stream/start`](#server--client-streamstart) or [`stream/clear`](#server--client-streamclear)) to the timestamp of the first subsequent audio chunk.
+  - `min_buffer_ms`: integer - requested minimum ongoing buffer duration in milliseconds during playback (primarily for live streams), used to absorb network jitter and ongoing decode/playback timing variance. Always required for players.
```
kahrendt, line 485: I wonder if there is a way we can give guidance on how to calculate this (`required_lead_time_ms` gives very specific instructions). This is a harder one to do reliably I guess, but if we could describe a way to approach this then that would be good.

I believe we had previously discussed something like looking at the distribution of time message latencies and taking like the 95th percentile or something along those lines. I think we'll have to do some experiments to really see what is actually usable in practice.

marcelveldt (approved):

PR #70, opened 2026-03-04
Add a color role

kahrendt:
Adds a new color role to the spec. This allows the server to send RGBA color data to clients for various purposes. It specifies background, foreground, accent, a light color for use on dark backgrounds, and a dark color use on light backgrounds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,5 +747,5 @@ The `color` object in [`server/state`](#server--client-serverstate) has this str
   - `is_dark?`: boolean | null - true if the background color is dark, suggesting light-colored content on top
   - `background?`: integer[] | null - background color as `[R, G, B]` with values 0-255
   - `accent?`: integer[] | null - accent color as `[R, G, B]` with values 0-255
-  - `light?`: integer[] | null - a light color derived from the source material, suitable for use on dark backgrounds, as `[R, G, B]` with values 0-255
-  - `dark?`: integer[] | null - a dark color derived from the source material, suitable for use on light backgrounds, as `[R, G, B]` with values 0-255
+  - `on_dark?`: integer[] | null - a light color derived from the source material, suitable for use on dark backgrounds, as `[R, G, B]` with values 0-255
+  - `on_light?`: integer[] | null - a dark color derived from the source material, suitable for use on light backgrounds, as `[R, G, B]` with values 0-255
--- a/README.md
+++ b/README.md
@@ -744,8 +744,8 @@ The `color` object in [`server/state`](#server--client-serverstate) has this str
 
 - `color`: object
   - `timestamp`: integer - server clock time in microseconds for when these colors are valid
-  - `is_dark?`: boolean | null - true if the background color is dark, suggesting light-colored content on top
-  - `background?`: integer[] | null - background color as `[R, G, B]` with values 0-255
+  - `background_dark?`: integer[] | null - background color suitable for dark mode as `[R, G, B]` with values 0-255
+  - `background_light?`: integer[] | null - background color suitable for light mode as `[R, G, B]` with values 0-255
   - `accent?`: integer[] | null - accent color as `[R, G, B]` with values 0-255
   - `on_dark?`: integer[] | null - a light color derived from the source material, suitable for use on dark backgrounds, as `[R, G, B]` with values 0-255
   - `on_light?`: integer[] | null - a dark color derived from the source material, suitable for use on light backgrounds, as `[R, G, B]` with values 0-255
--- a/README.md
+++ b/README.md
@@ -744,8 +744,8 @@ The `color` object in [`server/state`](#server--client-serverstate) has this str
 
 - `color`: object
   - `timestamp`: integer - server clock time in microseconds for when these colors are valid
-  - `background?`: integer[] | null - background color as `[R, G, B, A]` with values 0-255
-  - `foreground?`: integer[] | null - foreground/text color as `[R, G, B, A]` with values 0-255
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -731,3 +734,18 @@ Binary messages should be rejected if there is no active stream.
...
+
+The `color` object in [`server/state`](#server--client-serverstate) has this structure:
+
+- `color`: object
+  - `timestamp`: integer - server clock time in microseconds for when these colors are valid
+  - `background?`: integer[] | null - background color as `[R, G, B, A]` with values 0-255
+  - `foreground?`: integer[] | null - foreground/text color as `[R, G, B, A]` with values 0-255
+  - `accent?`: integer[] | null - accent color as `[R, G, B, A]` with values 0-255
+  - `light?`: integer[] | null - a light color derived from the source material, suitable for use on dark backgrounds, as `[R, G, B, A]` with values 0-255
+  - `dark?`: integer[] | null - a dark color derived from the source material, suitable for use on light backgrounds, as `[R, G, B, A]` with values 0-255
```
maximmaxim345: I'm having a hard time thinking of use cases of an alpha value, maybe we should drop it for simplicity?

README.md
```diff
@@ -731,3 +734,18 @@ Binary messages should be rejected if there is no active stream.
...
+This section describes messages specific to clients with the `color` role, which receive colors derived from the current audio. Colors may be extracted from album artwork, provided by the music source, or manually programmed by the server.
+
+### Server → Client: `server/state` color object
+
+The `color` object in [`server/state`](#server--client-serverstate) has this structure:
+
+- `color`: object
+  - `timestamp`: integer - server clock time in microseconds for when these colors are valid
+  - `background?`: integer[] | null - background color as `[R, G, B, A]` with values 0-255
+  - `foreground?`: integer[] | null - foreground/text color as `[R, G, B, A]` with values 0-255
```
maximmaxim345: Probably having RGB values for the foreground is a bit too custom for some UI clients?

What if we instead either have a boolean property `is_dark`? If its true the clients knows to use white text. If false it uses black instead.

Alternatively we could also have two `background`s, one `background_dark` and one `background_light` so each client can decide by itself based on whether its in dark or light mode.

Or a mix of both? have two backgrounds and a boolean property that says whether the server would prefer the client to go into dark or light mode.

README.md
maximmaxim345: The two background color variant would also work pretty well with the `light` and `dark` colors.

But we should probably rename them to `on_dark` and `on_light` to avoid confusion so its more clear when to use what.

README.md
kahrendt: I replaced ``foreground`` with ``is_dark`` boolean.

I don't know if we can easily generate a background for dark themes and light themes easily on the server side?

kahrendt: I dropped the ``is_dark`` boolean and added ``background_dark`` and ``background_light``. I didn't add a boolean for exposing the server's preference for a mode. I'm not sure if that's a useful feature or not, I had a hard time coming up with a use case where the client shouldn't be the one making the decision on their own, but I'm also not very creative!

README.md
```diff
@@ -731,3 +734,18 @@ Binary messages should be rejected if there is no active stream.
...
+This section describes messages specific to clients with the `color` role, which receive colors derived from the current audio. Colors may be extracted from album artwork, provided by the music source, or manually programmed by the server.
+
+### Server → Client: `server/state` color object
+
+The `color` object in [`server/state`](#server--client-serverstate) has this structure:
+
+- `color`: object
+  - `timestamp`: integer - server clock time in microseconds for when these colors are valid
+  - `background_dark?`: integer[] | null - background color suitable for dark mode as `[R, G, B]` with values 0-255
+  - `background_light?`: integer[] | null - background color suitable for light mode as `[R, G, B]` with values 0-255
```
maximmaxim345:
```suggestion
  - `background_dark?`: integer[] | null - background color suitable for dark mode as `[R, G, B]` with values 0-255. The server must ensure a minimum WCAG contrast ratio of 4.5:1 with white text and with `on_dark` (if also present).
  - `background_light?`: integer[] | null - background color suitable for light mode as `[R, G, B]` with values 0-255. The server must ensure a minimum WCAG contrast ratio of 4.5:1 with black text and with `on_light` (if also present).
```

Should we make our definition more strict with WCAG contrast ratios? With 4.5:1 the client can be sure that it can directly use the colors while still being accessible with normal sized text.
We already do something similar in the Music Assistant frontend. There we aim for 5:1. Only drawback is that the colors might get darker/lighter than whats actually in the cover art.

We could technically also add a `primary` color if thats a problem.

README.md
maximmaxim345: Agree, only idea I can think of is if you want to set light/dark mode across all your Sendspin device based on a time schedule.

But in that case we would need to require Sendspin clients to use the `is_dark` boolean for their system UI which introduces a lot of other problems.

kahrendt:
@maximmaxim345 I added a `primary` color field as well based on your last suggestion. I think users can use `primary` and `accent` colors for non-display purposes; e.g., light effects, and they can use the `background_*` and `on_*` fields for making a UI with text overlay. I think we've got our bases covered here now for the purposes that I can think of!

maximmaxim345 (approved):
Looks good!
Should cover pretty much all use cases now.
Only other addition i can think of is multiple accent colors? If visualizer clients want to cycle through more colors?
But I think having 3 colors are already enough:
- `primary`
- `secondary`
- and the `background_light` color that can also be used for light effects since it needs to be bright by definition

kahrendt:
Thanks!

PR #77, opened 2026-04-13
WIP: Add encryption and authentication support

maximmaxim345:
Rough draft of how encryption support could look like.
Related: https://github.com/Sendspin/backlog/issues/32
Still very WIP with a couple limitations and potential security vulnerabilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ For `pairing_method: 'button'`, the client must not respond with [`encryption/co
 
 **Reconnection (previously paired):** The client includes its stored copy of the server's static public key in [`client/hello`](#client--server-clienthello). If the server recognizes the client, both sides perform a Noise KK handshake using stored static keys — one round trip, mutual cryptographic authentication, no user interaction.
 
-**Prologue binding:** For all Noise handshakes (NNpsk0 and KK), both sides must set the Noise handshake prologue to the concatenation of the raw `client/hello` and `server/hello` WebSocket text frame payloads, in that order. Each side uses the bytes it sent for its own message and the bytes it received for the peer's message. This binds the plaintext negotiation to the handshake — if an attacker modifies either hello message (e.g., changing the cipher suite), the handshake fails.
+**Prologue binding:** For all Noise handshakes (NNpsk0 and KK), both sides must set the Noise handshake prologue to the concatenation of the raw `client/hello` and `server/hello` WebSocket text frame payloads, in that order. Each side uses the bytes it sent for its own message and the bytes it received for the peer's message. Implementations must retain the exact bytes as sent and received; re-serializing parsed JSON is not guaranteed to produce identical bytes and will break the handshake. This binds the plaintext negotiation to the handshake — if an attacker modifies either hello message (e.g., changing the cipher suite), the handshake fails.
 
 **Initiator/responder:** The Sendspin client is always the Noise initiator and the Sendspin server is always the Noise responder, regardless of which side initiated the WebSocket connection.
 
--- a/README.md
+++ b/README.md
@@ -97,6 +97,58 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
 
 While custom connection methods are possible for specialized use cases (like remotely accessible web-browsers, mobile apps), most clients should use one of the two standardized methods above if possible.
 
+## Encryption
+
+Sendspin uses the [Noise Protocol Framework](https://noiseprotocol.org/) to protect communication between clients and servers. Encryption guards against rogue devices injecting commands or malicious data, and prevents eavesdropping on audio streams and metadata on the local network.
+
+Servers must implement encryption. Clients should implement encryption. Whether a server allows unencrypted clients is application-specific and not defined by this specification.
+
+Encryption operates at the message level. The WebSocket connection itself remains unencrypted — individual message payloads are encrypted after the Noise handshake completes. This allows the [`client/hello`](#client--server-clienthello) and [`server/hello`](#server--client-serverhello) handshake to remain plaintext on the same connection and provides compatibility with WebSocket proxies and load balancers.
+
+### Cipher Suites
+
+This specification defines two cipher suites. Servers must support both. Clients select their preferred suite in [`client/hello`](#client--server-clienthello) based on their hardware capabilities.
+
+- `25519_ChaChaPoly_SHA256` — Curve25519, ChaCha20-Poly1305, SHA-256. Fast in software, recommended for embedded devices without hardware AES (e.g., ESP32).
+- `25519_AESGCM_SHA256` — Curve25519, AES-256-GCM, SHA-256. Benefits from hardware AES acceleration on most modern CPUs, phones, and set-top boxes (AES-NI, ARM CE).
+
+### Pairing and Reconnection
+
+There are two encryption flows depending on whether the client has previously paired with this server.
+
+**Pairing (first connection):** The server presents the new client in its UI for user approval. Once approved, both sides perform an ephemeral Diffie-Hellman key exchange to derive a pre-shared key (PSK) without ever sending it over the wire:
+
+```
+PSK = HKDF-SHA256(dh_shared_secret, salt="sendspin-pairing", info="sendspin-psk")
+```
+
+This PSK bootstraps a Noise NNpsk0 handshake. Inside the resulting encrypted session, both sides exchange long-term static public keys for future reconnections. See [encryption messages](#encryption-messages) for the full message sequence.
+
+**Reconnection (previously paired):** The client includes its stored copy of the server's static public key in [`client/hello`](#client--server-clienthello). If the server recognizes the client, both sides perform a Noise KK handshake using stored static keys — one round trip, mutual cryptographic authentication, no user interaction.
+
+**Prologue binding:** For all Noise handshakes (NNpsk0 and KK), both sides must set the Noise handshake prologue to the concatenation of the raw `client/hello` and `server/hello` WebSocket text frame payloads, in that order. Each side uses the bytes it sent for its own message and the bytes it received for the peer's message. This binds the plaintext negotiation to the handshake — if an attacker modifies either hello message (e.g., changing the cipher suite), the handshake fails.
+
+**Initiator/responder:** The Sendspin client is always the Noise initiator and the Sendspin server is always the Noise responder, regardless of which side initiated the WebSocket connection.
+
+If the server does not recognize the client (e.g., the server's key was rotated or the client was revoked), the server responds with `paired: false` and the client falls back to the pairing flow.
+
+### Encrypted Message Format
+
+Once the Noise handshake completes, all WebSocket messages are sent as binary frames containing Noise ciphertext. The reserved binary message ID `0x00` distinguishes encrypted text messages from encrypted binary messages:
+
+- **Encrypted text message:** `0x00` + JSON bytes → encrypt → send as WebSocket binary frame
+- **Encrypted binary message:** original binary payload (existing type byte + data) → encrypt → send as WebSocket binary frame
+
+The receiver decrypts and checks the first byte: `0x00` means strip it and parse the rest as JSON; any other value is handled as a binary message using the existing [binary message ID structure](#binary-message-id-structure).
+
+Nonce management is handled automatically by the Noise transport state.
+
+### Key Storage
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,14 +97,68 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
 
 While custom connection methods are possible for specialized use cases (like remotely accessible web-browsers, mobile apps), most clients should use one of the two standardized methods above if possible.
 
+## Encryption
+
+Sendspin uses the [Noise Protocol Framework](https://noiseprotocol.org/) to protect communication between clients and servers. Encryption guards against rogue devices injecting commands or malicious data, and prevents eavesdropping on audio streams and metadata on the local network.
+
+Servers must implement encryption. Clients should implement encryption. Whether a server allows unencrypted clients is application-specific and not defined by this specification.
+
+Encryption operates at the message level. The WebSocket connection itself remains unencrypted — individual message payloads are encrypted after the Noise handshake completes. This allows the [`client/hello`](#client--server-clienthello) and [`server/hello`](#server--client-serverhello) handshake to remain plaintext on the same connection and provides compatibility with WebSocket proxies and load balancers.
```
maximmaxim345, line 108: An active MITM can strip the encryption field from the plaintext `client/hello`, causing the server to believe the client doesn't support encryption. 
We should probably require encryption for all clients, or at least define that servers should reject unencrypted clients by default.
Sendspin servers could then have a advanced option to allow insecure clients.

README.md
```diff
@@ -97,14 +97,68 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+
+There are two encryption flows depending on whether the client has previously paired with this server.
+
+**Pairing (first connection):** The server presents the new client in its UI for user approval. Once approved, both sides perform an ephemeral Diffie-Hellman key exchange to derive a pre-shared key (PSK) without ever sending it over the wire:
+
+```
+PSK = HKDF-SHA256(dh_shared_secret, salt="sendspin-pairing", info="sendspin-psk")
+```
+
+This PSK bootstraps a Noise NNpsk0 handshake. Inside the resulting encrypted session, both sides exchange long-term static public keys for future reconnections. See [encryption messages](#encryption-messages) for the full message sequence.
```
maximmaxim345, line 134: This version of the spec does not secure against MITM attacks.
We need to support mandatory PIN confirmation or similar, at least for commercial devices. Probably still optional for DIY clients.

README.md
```diff
@@ -97,14 +97,68 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+
+**Prologue binding:** For all Noise handshakes (NNpsk0 and KK), both sides must set the Noise handshake prologue to the concatenation of the raw `client/hello` and `server/hello` WebSocket text frame payloads, in that order. Each side uses the bytes it sent for its own message and the bytes it received for the peer's message. This binds the plaintext negotiation to the handshake — if an attacker modifies either hello message (e.g., changing the cipher suite), the handshake fails.
+
+**Initiator/responder:** The Sendspin client is always the Noise initiator and the Sendspin server is always the Noise responder, regardless of which side initiated the WebSocket connection.
+
+If the server does not recognize the client (e.g., the server's key was rotated or the client was revoked), the server responds with `paired: false` and the client falls back to the pairing flow.
```
maximmaxim345: An attacker could forge `server/hello` with `paired: false`, forcing the client into the pairing flow. The client sends `server_static_key` in plaintext, so the attacker doesn't need the server's private key.

What could help against this is:
- Require clients to disconnect instead of silently falling back to pairing. In this case clients need some sort of paring mode.
- Maybe send a cryptographic hash of the server static key instead of the full key?

README.md
```diff
@@ -97,14 +97,68 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+### Encrypted Message Format
+
+Once the Noise handshake completes, all WebSocket messages are sent as binary frames containing Noise ciphertext. The reserved binary message ID `0x00` distinguishes encrypted text messages from encrypted binary messages:
+
+- **Encrypted text message:** `0x00` + JSON bytes → encrypt → send as WebSocket binary frame
+- **Encrypted binary message:** original binary payload (existing type byte + data) → encrypt → send as WebSocket binary frame
+
+The receiver decrypts and checks the first byte: `0x00` means strip it and parse the rest as JSON; any other value is handled as a binary message using the existing [binary message ID structure](#binary-message-id-structure).
+
+Nonce management is handled automatically by the Noise transport state.
```
maximmaxim345, line 163: Noise has a 65535-byte transport message limit. Large artwork images may exceed this.

README.md
```diff
@@ -97,14 +97,68 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+
+### Key Storage
+
+Clients must support storing at least 5 server entries, using LRU (least recently used) eviction when the limit is reached. Each entry contains the `server_id` and the server's long-term static public key. The client also stores its own long-term static keypair (one keypair, used with all servers). All keys must persist across reboots.
+
+Servers store all approved clients with no limit defined by this specification. Each entry contains the `client_id` and the client's long-term static public key. The server also stores its own long-term static keypair. All keys must persist across reboots. Servers should provide a way for users to revoke individual clients, removing their stored key and forcing re-pairing.
```
maximmaxim345, line 169: LRU eviction of stored server keys could be exploited by malicious servers to fill all 5 slots, forcing the client to re-pair with a legitimate server. This re-pairing could enable a MITM attack since the user must re-approve the client.

A explicit paring mode for clients could help against this issue too.

README.md
```diff
@@ -435,6 +516,49 @@ Upon receiving this message, the server should initiate the disconnect.
 
 **Note:** Clients may close the connection without sending this message (e.g., crash, network loss), or immediately after sending `client/goodbye` without waiting for the server to disconnect. When a client disconnects without sending `client/goodbye`, servers should assume the disconnect reason is `restart` and attempt to auto-reconnect.
 
+## Encryption messages
+This section describes messages used to establish an encrypted session between client and server. These messages are exchanged after the [`client/hello`](#client--server-clienthello) / [`server/hello`](#server--client-serverhello) handshake and before any other communication. All encryption messages are sent as plaintext JSON, except [`encryption/keys`](#bidirectional-encryptionkeys) which is the first encrypted message after the Noise handshake completes.
+
+### Server → Client: `encryption/approval`
```
maximmaxim345, line 543: Maybe we should add a timeout behavior during pairing is undefined. For server-initiated connections, a server that never approves or denies could block the client from connecting to other servers.

We definitely need to think more about how we handle multi-server scenarios.

README.md
maximmaxim345: Resolved. Servers are now required to dismiss unencrypted clients by default.

README.md
maximmaxim345: Revolved. Servers are required to warn the user before re-paring.

README.md
maximmaxim345: Partially resolved. LRU eviction requires a successful pairing, which requires a physical button press on clients that support `pairing_method: 'button'`.

README.md
maximmaxim345: Resolved. Pairing slot is separate from active connection. No need for a timeout anymore.

README.md
maximmaxim345: PIN is difficult though, only very few speakers have screens.

We could also just accept the risk since the MITM attack needs to happen during setup. After that the connection is protected.
And for devices that implement `pairing_method: 'button'` the LRU eviction issue can only happen with physical access.

README.md
```diff
@@ -97,14 +99,85 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+This specification defines two cipher suites. Servers must support both. Clients select their preferred suite in [`client/hello`](#client--server-clienthello) based on their hardware capabilities.
+
+- `25519_ChaChaPoly_SHA256` — Curve25519, ChaCha20-Poly1305, SHA-256. Fast in software, recommended for embedded devices without hardware AES (e.g., ESP32).
+- `25519_AESGCM_SHA256` — Curve25519, AES-256-GCM, SHA-256. Benefits from hardware AES acceleration on most modern CPUs, phones, and set-top boxes (AES-NI, ARM CE).
+
+### Pairing Methods
+
+Clients advertise their pairing capability via the `pairing_method` field in [`client/hello`](#client--server-clienthello):
+
+- `button` — physical pairing button on the device. Pairing requires both server UI approval and a button press. **Recommended.**
```
maximmaxim345, line 121: What is that button? play/pause? Long pressing play/pause? Dedicated sync/paring button?

Not clear.

README.md
```diff
@@ -275,10 +367,16 @@ Players that can output audio should have the role `player`.
   - `metadata@v1` - displays text metadata describing the currently playing audio
   - `artwork@v1` - displays artwork images
   - `visualizer@v1` - visualizes audio
+- `encryption?`: object - omitted if client does not support [encryption](#encryption)
+  - `suite`: string - preferred cipher suite: `'25519_ChaChaPoly_SHA256'` or `'25519_AESGCM_SHA256'`
+  - `pairing_method`: string - `'button'` or `'none'`. See [pairing methods](#pairing-methods)
+  - `server_static_key?`: string - Base64 encoded static public key of this server from a previous pairing. Enables [reconnection](#pairing-and-reconnection) via Noise KK handshake
```
maximmaxim345, line 373: The client doesn't know yet what server is connecting. How can then the client send the correct static key?

But thinking about it, we don't even need this field. The server and client can just try to build a connection, if it fails we treat it invalid.
And if the client suddenly acts as if it isn't paired, the server should also show a warning.

README.md
Hedda: Remember that not all speakers have a buttons though, (so would be nice if could force pairing without a button). 

I guess an alternative option to having a physical pairing button could be to do what many Zigbee devices manufactures do when the device does not have any buttons and instead have the user power ON and OFF the device a certain amout of times within a set time slot? The downside to that is that it is not a user-friendly experience and pairing of some Zigbee devices that uses that method can sometimes require multiple or repeated attempts before pairing works, (that is also why many manufacturers usually recommend to always factory-reset of such Zigbee devices before pairing it).

maximmaxim345: Yep, this is a tricky problem to solve though.

Power cycling wouldn't really work here since you need to keep the WebSocket connection during this "accept pairing" action. (and I hate switching the power on and off every time I need to pair a lightbulb haha)

As an alternative there is still the `none` paring option. But here its a compromise between security and convenience. For example, anybody with access to the same network can blast music with full volume.

Another idea I'm thinking about is to have another `password` pairing option. Here any server that wants to pair to the server needs to know either a factory provided password (thats maybe printed on a sticker).
Or a device without a button or provided password can be "upgraded" to this more secure form.

I think most users wouldn't set a password, but with the option you could setup any Sendspin client (even without a button) to be secure enough to be used on a public wifi (for example in a Cafe or something similar). At least when just looking at the protocol itself as a attack surface.

I'm pretty sure this password approach is used by at least a couple AirPlay speakers.

README.md
kahrendt: Reserve a bit of the role byte. a 0 (existintg behavior) says its a finished message. A 1 says its unfinished and will continue with follow up messages until one is marked finished.

README.md
```diff
@@ -97,14 +99,85 @@ Unlike server-initiated connections, servers cannot reclaim clients by reconnect
...
+
+- `25519_ChaChaPoly_SHA256` — Curve25519, ChaCha20-Poly1305, SHA-256. Fast in software, recommended for embedded devices without hardware AES (e.g., ESP32).
+- `25519_AESGCM_SHA256` — Curve25519, AES-256-GCM, SHA-256. Benefits from hardware AES acceleration on most modern CPUs, phones, and set-top boxes (AES-NI, ARM CE).
+
+### Pairing Methods
+
+Clients advertise their pairing capability via the `pairing_method` field in [`client/hello`](#client--server-clienthello):
+
+- `button` — physical pairing button on the device. Pairing requires both server UI approval and a button press. **Recommended.**
+- `none` — no physical pairing mechanism. Pairing requires only server UI approval. **Discouraged** — vulnerable to rogue server pairing with no physical presence check.
```
maximmaxim345, line 122: Alternative idea to always require the button press:
Clients allow all servers to connect, but some roles require a more secure paring process.

In that case playing to a speaker is always as frictionless as possible (and still protects the server from zero-click attacks).
But more sensitive roles are still secured behind additional steps (for example the future source role).

For example, a client implementing the `player` and `source` role will only be able to use the `player` role until the user does the full pairing process.

maximmaxim345:
Closing since the development of encryption support is continued in:
- #84

Issue #79, opened 2026-04-26
Feature request - add Lyrics role


OnFreund:
Add a Lyrics role so SendSpin servers can share track lyrics with clients, similar to the Artwork role.
The support object will include an ordered lists of requested formats.

Open questions:
1. Should we have `lyrics_url` as part of the metadata similar to artwork, or just binary support?
2. What lyrics formats should be supported?
3. Should we have one lyrics stream per track (like audio), or one per channel (like artwork)? The latter could be useful in case multiple formats are available (e.g. both karaoke and text)

balloob:
Not sure about URL, I think we can just provide it. I do like the channel idea, 

Should we have some switch command because you don't want them all? How far are we going with this. Do we support subtitles for audio books in other languages ? And if it's a 10 hour audio book, the lyrics can't all be sent at once (this is probably too big an edge case to cover).

Probably MVP keep it simple, 1 format, no channels?

OnFreund:
Audiobooks are a great question, though even a 10 hour audio book would be relatively small if it's text only (the entire LOTR trilogy should be about 3MB in UTF-8 - should be ok over a local network). I think we can probably send at once, but add a size limit.

I do think that karaoke subtitles are going to become very popular, so I would consider 2 channels, and support `.lrc` and `.cdg`.

Hedda:
Directly related to https://github.com/Sendspin/spec/issues/1 or duplicate?

OnFreund:
Thanks @Hedda, not sure how I missed it 🤦‍♂️

Hedda:
I understand that [LRC (text-based)](https://en.wikipedia.org/wiki/LRC_(file_format)) and [CDG (graphics-based)](https://en.wikipedia.org/wiki/CD%2BG) karaoke lyrics files are the two most common karaoke formats today, 

* You can find some more good info formatting LRC here -> https://kodi.wiki/view/Archive:Create_LRC_karaoke_lyrics_files
* As relevant to Sendspin and Music Assistant the CDG format is primarly available as MP3+CDG (MP3 + file with timed graphics).

Another relativly common karaoke format is KAR which more or less contain Karaoke Midi Music + timed lyrics in text format. 

> do think that karaoke subtitles are going to become very popular, so I would consider 2 channels, and support `.lrc` and `.cdg`.

[...]

PR #80, opened 2026-04-26
feat: add Lyrics role (LRC and CDG)

OnFreund:
Implements Sendspin/spec#79.

## Summary

- Adds a new `lyrics@v1` role for clients that display track lyrics or karaoke subtitles
- Supports two channels (0–1) and two formats: `lrc` (text with embedded timestamps) and `cdg` (CD+G binary karaoke)
- Lyrics are delivered as binary messages (types 12–13, assigned from the previously reserved `000011xx` slot), without transport timestamps since timing is embedded in the files themselves
- Clients can toggle channels on/off via `stream/request-format` without reconnecting

## Details

- `client/hello`: new `lyrics@v1_support` object with `channels[]` declaring format per channel (`lrc` | `cdg` | `none`)
- `stream/start`: new `lyrics` object confirming active channel formats
- `stream/request-format`: new `lyrics` object for mid-session format/channel changes
- `stream/end`: `lyrics` added to the roles array
- Binary message IDs 12–13 assigned to Lyrics role
- Clearing lyrics: empty binary message (type byte only, no payload)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ sequenceDiagram
             Server->>Client: binary Types 8-11 (artwork channels 0-3)
         end
         alt Lyrics role
-            Server->>Client: binary Types 12-13 (lyrics channels 0-1)
+            Server->>Client: binary Type 12 (lyrics data)
         end
         alt Visualizer role
             Server->>Client: binary Type 16 (visualization data)
@@ -714,8 +714,6 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
 ## Lyrics messages
 This section describes messages specific to clients with the `lyrics` role, which handle display of track lyrics and karaoke subtitles. Lyrics clients receive lyrics data in their preferred format delivered as binary messages.
 
-**Channels:** Lyrics clients can support 1-2 independent channels. This allows a client to simultaneously receive multiple lyrics formats (e.g., plain LRC text on one channel and CDG karaoke data on another). Each channel operates independently with its own format.
-
 **Supported formats:**
 - `lrc` - LRC text format with embedded timestamps (UTF-8 encoded)
 - `cdg` - CD+G karaoke format (binary)
@@ -727,10 +725,8 @@ This section describes messages specific to clients with the `lyrics` role, whic
 The `lyrics@v1_support` object in [`client/hello`](#client--server-clienthello) has this structure:
 
 - `lyrics@v1_support`: object
-  - `channels`: object[] - list of supported lyrics channels (length 1-2), array index is the channel number
-    - `format`: 'lrc' | 'cdg' | 'none' - requested lyrics format
-
-**None format:** If a channel has `format` set to `none`, the server will not send any lyrics data for that channel. This allows clients to disable and enable specific channels on the fly through [`stream/request-format`](#client--server-streamrequest-format-lyrics-object) without needing to re-establish the WebSocket connection.
+  - `supported_formats`: string[] - list of supported lyrics formats in priority order (first is preferred), e.g. `['lrc', 'cdg']`
+  - `max_size_bytes`: integer - maximum size in bytes of lyrics data the client can handle per binary message. If the lyrics for the current track exceed this limit, the server must send [`stream/end`](#server--client-streamend) for the lyrics role instead of the binary data.
 
 **Note:** Servers must support all lyrics formats: 'lrc' and 'cdg'.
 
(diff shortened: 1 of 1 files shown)


OnFreund:
Note - even though we wanted to send the lyrics inline, I looked into CDG files and they can be quite large, so I went with binary messages after all. This also keeps it consistent with artwork.

README.md
```diff
@@ -700,6 +708,64 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
 **Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty binary message (only the message type byte and timestamp, with no image data) for that channel.
 
+## Lyrics messages
+This section describes messages specific to clients with the `lyrics` role, which handle display of track lyrics and karaoke subtitles. Lyrics clients receive lyrics data in their preferred format delivered as binary messages.
+
+**Channels:** Lyrics clients can support 1-2 independent channels. This allows a client to simultaneously receive multiple lyrics formats (e.g., plain LRC text on one channel and CDG karaoke data on another). Each channel operates independently with its own format.
+
+**Supported formats:**
+- `lrc` - LRC text format with embedded timestamps (UTF-8 encoded)
+- `cdg` - CD+G karaoke format (binary)
```
maximmaxim345, line 719: lrc files are pretty ubiquitous, but I haven't heard of CD+G before.
Why should we support CD+G? I mean if it's really niche we can't depend on Sendspin implementations supporting that.

README.md
```diff
@@ -700,6 +708,64 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+
+**No transport timestamps:** Unlike audio or artwork, lyrics binary messages do not carry a server clock timestamp. Timing information is embedded within the lyrics data itself (e.g., `[mm:ss.xx]` tags in LRC, frame timing in CDG). The server sends lyrics at the start of each track; clients begin processing as soon as they receive them.
```
maximmaxim345, line 721: How will the lyrics be in sync then if it's not aligned to any timestamp?
I mean:
- What if you pause/stop playback in the mean time, the lyrics should also stop in that case
- Depending on the networks latency, the position in the lyrics can be different every time.

README.md
```diff
@@ -700,6 +708,64 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+
+### Client → Server: `client/hello` lyrics@v1 support object
+
+The `lyrics@v1_support` object in [`client/hello`](#client--server-clienthello) has this structure:
+
+- `lyrics@v1_support`: object
+  - `channels`: object[] - list of supported lyrics channels (length 1-2), array index is the channel number
```
maximmaxim345: Why would we want two lyrics channels visible at once?
I think we can simplify this to remove the channels concept here.

README.md
```diff
@@ -700,6 +708,64 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+- Byte 0: message type `12`-`13` (uint8) - corresponds to lyrics channel 0-1 respectively
+- Rest of bytes: lyrics data (UTF-8 encoded text for `lrc`; binary data for `cdg`)
+
+The message type determines which lyrics channel this data is for:
+- Type `12`: Channel 0 (Lyrics role, slot 0)
+- Type `13`: Channel 1 (Lyrics role, slot 1)
+
+The server sends lyrics data at the start of each track and resends after a track change. Clients begin processing the lyrics data immediately upon receipt.
+
+**Clearing lyrics:** To indicate that no lyrics are available for the current track on a specific channel, the server sends an empty binary message (only the message type byte, with no lyrics data) for that channel.
```
maximmaxim345: I think we can also use `stream/stop` for that.

README.md
```diff
@@ -700,6 +708,64 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+- `lyrics`: object
+  - `channels`: object[] - configuration for each active lyrics channel, array index is the channel number
+    - `format`: 'lrc' | 'cdg' | 'none' - format of the lyrics data
+
+### Server → Client: Lyrics (Binary)
+
+Binary messages should be rejected if there is no active stream.
+
+- Byte 0: message type `12`-`13` (uint8) - corresponds to lyrics channel 0-1 respectively
+- Rest of bytes: lyrics data (UTF-8 encoded text for `lrc`; binary data for `cdg`)
```
maximmaxim345, line 756: (How) Should we handle low memory clients?
For timed lyrics we could think of something fancy like only send x seconds of lyrics in advance, like we do now with audio.

But on the other hand that will lock the client out of supporting non timed lyrics, at least in some edge cases. (and requires the clients to implement our own format instead of being able to parse standard lrc files).

So, to keep it simple maybe it's worth having a hard byte limit set by the client?

maximmaxim345: `CD+G` is definitely even trickier since it's almost like an uncompressed video file.

README.md
OnFreund: I don't have the data to back this up unfortunately, but karaoke subtitles are not niche and are well supported by both hardware and software.

README.md
OnFreund: The spec instructs clients how to calculate track progress - can't we use that progress value to show the right lyrics? This is what I'm currently doing in my client (which is downloading lyrics independently) and it's working well.

README.md
OnFreund: Technically we don't need it at once, but the server might have just one kind for a track. I guess we can request both and show what the server is responding with.

README.md
Hedda: > lrc files are pretty ubiquitous, but I haven't heard of CD+G before.
> Why should we support CD+G? I mean if it's really niche we can't depend on Sendspin implementations supporting that.

@maximmaxim345 CDG+G is not niche at all, it is an official standard for Audio-CD with Graphics (as Audio Compact Discs with digital images) that all professional Karaoke machines uses (such as used in Karaoke bars and home use), so almost all songs are available in that format.The ”G” stands for Graphics and as I understand the format is it just images for each character and as such does not need a font engine to display wierd characters for all languages. Anyway, more common nowadays are MP3+G which is just when an the redbook Audio-CD part been converted to MPG and the Graphics broken out to separate files (.cdg).

* https://en.wikipedia.org/wiki/CD%2BG

image

For player code reference I suggest check out the Kodi (formerly XBMC) codebase which supports CD+G and MP3+G

* https://github.com/xbmc/xbmc

Quote from that Wikipedia article:

Along with dedicated karaoke machines, other consumer devices that play CD+G format CDs include the NEC [TurboGrafx-CD](https://en.wikipedia.org/wiki/TurboGrafx-CD) (a CD-ROM peripheral for the [TurboGrafx-16](https://en.wikipedia.org/wiki/TurboGrafx-16)) and [Turbo Duo](https://en.wikipedia.org/wiki/Turbo_Duo), as well as the Japan-only successor the [PC-FX](https://en.wikipedia.org/wiki/PC-FX), the [Philips CD-i](https://en.wikipedia.org/wiki/Philips_CD-i), the [Sega CD](https://en.wikipedia.org/wiki/Sega_CD), [Sega Saturn](https://en.wikipedia.org/wiki/Sega_Saturn),[[5]](https://en.wikipedia.org/wiki/CD%2BG#cite_note-6) the JVC [X'Eye](https://en.wikipedia.org/wiki/X%27Eye), the [3DO Interactive Multiplayer](https://en.wikipedia.org/wiki/3DO_Interactive_Multiplayer), the [Amiga CD32](https://en.wikipedia.org/wiki/Amiga_CD32) and [Commodore CDTV](https://en.wikipedia.org/wiki/Commodore_CDTV), and the [Atari Jaguar CD](https://en.wikipedia.org/wiki/Atari_Jaguar_CD) (an attachment for the [Atari Jaguar](https://en.wikipedia.org/wiki/Atari_Jaguar)). Some CD-ROM drives can also read this data. Pioneer's [LaserActive](https://en.wikipedia.org/wiki/LaserActive) player can also play CD+G discs, as long as either the PAC-S1/S-10 or PAC-N1/N10 game modules are installed.

Since 2003, some standalone [DVD players](https://en.wikipedia.org/wiki/DVD_player) have supported the CD+G format. Regular [audio CD players](https://en.wikipedia.org/wiki/CD_player) will output only the audio tracks as if it was a normal music CD, unless otherwise designed to read the extra data (lyrics and images).[[6]](https://en.wikipedia.org/wiki/CD%2BG#cite_note-7)

CD+G karaoke albums are still made today by several UK and US manufacturers including Sunfly, Zoom Entertainments, SBI Karaoke and Vocal Star. Although the popularity of CD sales are dwindling the format is still widely used as [MP3+G](https://en.wikipedia.org/wiki/MP3%2BG) downloads.

Hedda:
@maximmaxim345 any chance this Lyrics role `could/will make it into the the final Sendspin v1 (1.0 version) specification?

maximmaxim345:
> @maximmaxim345 any chance this Lyrics role `could/will make it into the the final Sendspin v1 (1.0 version) specification?

No, it won't make it into v1. But since Sendspin will be backwards compatible we don't need a new major Sendspin version to add a new role like lyrics.

There are still some issues with this PR I'm not fully convinced yet, those just need a bit more work to get ironed out before it's ready.

I haven't had time to look into this PR recently, but the main issues from what I remember were:
- CD+G is IMO still a legacy format that complicates server and client implementations considerably. (looks like it's getting replaced by proprietary formats and mp4 during the past 2 decades.
- Having both CD+G and lrc support will probably causes lrc clients to just not show anything while only the hand full of cdg compatible clients can display it, unless:
  - We force the server to convert cdg on the fly to lrc - in this case why even support cd+g in the first place?
  - We explain users why some clients work while others don't. This is a bad UX, especially if the server can't convert lrc to CD+G.
  - We force all clients to implement the CD+G spec, which is IMO not worth the complexity increase
- It might be advantageous to use a custom messages instead of lrc if that avoids the memory issues on long tracks and improves handling on track transitions. Servers need to already replace timestamps in lrc files after all. This will also cut down the overhead when seeking since we don't need to send the whole tracks lyrics at once. < I haven't looked into this in more detail yet though.

PR #81, opened 2026-04-27
Move repeat and shuffle states to controller state

kahrendt:
These two states seem out of place for the metadata role, as they don't describe more textual details about the current playing audio. They do describe how the audio will playback (much like how volume and mute states describe it), and they are group wide settings. This PR moves those two fields to the controller state and makes them always required fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,6 +593,8 @@ The `controller` object in [`server/state`](#server--client-serverstate) has thi
   - `supported_commands`: string[] - subset of: 'play' | 'pause' | 'stop' | 'next' | 'previous' | 'volume' | 'mute' | 'repeat_off' | 'repeat_one' | 'repeat_all' | 'shuffle' | 'unshuffle' | 'switch'
   - `volume`: integer - volume of the whole group, range 0-100
   - `muted`: boolean - mute state of the whole group
+  - `repeat`: 'off' | 'one' | 'all' - repeat mode: 'off' = no repeat, 'one' = repeat current track, 'all' = repeat all tracks (in the queue, playlist, etc.)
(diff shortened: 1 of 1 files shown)


kahrendt:
Metadata describes what is playing. Controller describes how it is playing.

maximmaxim345:
> Metadata describes what is playing. Controller describes how it is playing.

Makes sense! At least more sense than what it was before.
Should we add this to the Spec as well then? So we don't forget how to split "metadata" between both roles if we ever need to add more fields? But this PR might already be good enough as a documentation for the reasoning.

The only other consistent alternative I can think of that is putting everything into the `metadata` state.
But then you have a `controller` role where you can only write the group volume without ever knowing what the actual volume is right now. That sounds even more backwards.

maximmaxim345 (approved):

chrisuthe:
I agree, this is the right way to approach, I think previously the spec awkwardly tried to keep the controller role mostly focused on the command behavior, and less about getting data back from the Server. There isn't a reason that existing split should be continued imho.  We do need to update the spec + website though, to match.

kahrendt:
Thanks!

> Makes sense! At least more sense than what it was before. Should we add this to the Spec as well then? So we don't forget how to split "metadata" between both roles if we ever need to add more fields? But this PR might already be good enough as a documentation for the reasoning.

I didn't see a great spot for adding that to the spec in general, we don't really have a pattern for the "why" questions in the spec itself. Maybe we should consider adding some supplemental document that we can update that has more context on how/why we made certain decisions.

@chrisuthe Would you mind handling the website update? I think there are now a few things out of date (a couple weeks ago didn't we discuss that the pause state was still there? Maybe that's been fixed though)

PR #83, opened 2026-05-02
Clarify use of `stream/start`, `stream/end` and `stream/clear` messages

OnFreund:
Clarify the conditions under which the `stream/end` message is sent.

This change makes the current implementation of Music Assistant non-compliant, but we ensure that other servers don't copy this behavior, which forces unneeded complexity on clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,10 @@ sequenceDiagram
         Server->>Client: stream/clear (roles: [player, visualizer])
     end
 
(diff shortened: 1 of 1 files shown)


kahrendt:
Thanks! We were held back only because the VPE firmware had a bug where this was broken, but stable release 26.6.0 has the proper fix and this can proceed.

PR #84, opened 2026-05-13
Add encryption support

arturpragacz:
Add encryption support


README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+  - `concurrent_attempt` - the client refused the connection because a higher-or-equal-priority connection is already active (e.g., a `'management'` session, or a pairing handshake when the incoming reason is also a pairing). Server may retry later
+  - `unpaired` - the client has processed [`server/unpair`](#server--client-serverunpair) from this server. Server should not auto-reconnect
+
+**Note:** Clients may close the connection without sending this message (e.g., crash, network loss), or immediately after sending `client/goodbye` without waiting for the server to disconnect. When a client disconnects without sending `client/goodbye`:
+
+- On a `'playback'` or `'discovery'` connection, servers should assume the disconnect reason is `restart` and attempt to auto-reconnect.
+- On a `'pairing'` or `'management'` connection, servers should treat the drop as a session termination and not auto-reconnect; resumption, if desired, is operator-driven.
+- Servers should also apply backoff on repeated Noise-handshake failures to avoid tight reconnect loops when a long-term PSK has become invalid (e.g., after a client factory reset).
+
+## Pairing
```
maximmaxim345, line 638: To play music a client needs to be paired. To pair the user has to enter a PIN or a 43 character code into the server.

With that we force the user to find a sticker or type in a code when the user just wants to play music. The primary thing that we want to protect is the server from malicious actors, so I'm not sure it's worth it passing that burden onto the users to hunt for a sticker.
The worst case UX I can imagine is:
- The speaker didn't initially ship with Sendspin since that was only added after an update. -> Therefore there is no pairing sticker on the device.
- With that, the user needs to research the internet where to find the pairing code. -> Apparently its in the App
- The user install the Speaker manufacturers App and needs to login there.
- The user has to find the Sendspin pairing pins of that speaker somewhere in the App
- And has to type that into a PC running the Sendspin Server. This is even worse if the speaker only supports PSK, meaning he needs to manually type the 43 character code.

maximmaxim345: I like Paulus idea only requiring the pairing process with sensitive roles.
So regular playback through the `player` role can stay as simple as possible, while the `source` role requires you to pair first for added security.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+2. **Dynamic PIN** - pairing with a per-session [Sendspin Pairing PIN](#definitions); the client derives the PIN from a commit-and-reveal binding to the Noise handshake and emits it via an out-channel (display, speaker, etc.) for the operator to enter into the server. See [Dynamic PIN Pairing Flow](#dynamic-pin-pairing-flow).
+3. **Static PIN** - pairing with a fixed [Sendspin Pairing PIN](#definitions). Appropriate for devices with no out-channel; vulnerable to MITM if the PIN is disclosed. See [Static PIN Pairing Flow](#static-pin-pairing-flow).
+
+Static pairing methods (Pairing PSK, static PIN) do not take over the device's out-channel. Dynamic pairing (dynamic PIN) takes over the out-channel - typically the audio output or display - to emit the per-session PIN, so it cannot run while audio is playing on the same device. All pairing handshakes displace any active playback for their duration; see [Multiple servers](#multiple-servers).
+
+Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pin` rather than `static_pin`. `static_pin` is intended only for devices that genuinely cannot emit a per-session value.
```
maximmaxim345, line 652: Reading an 8-digit PIN aloud isn't really usable. Languages complicate it you'll need TTS or pre-recorded samples for every supported locale.
But also with just english Resource constrained devices can't ship a TTS stack or pre-recorded sample banks that adds MBs of flash.

README.md
```diff
@@ -90,22 +94,101 @@ Clients discover the server through mDNS and initiate a WebSocket connection usi
...
 
 Unlike server-initiated connections, servers cannot reclaim clients by reconnecting. How clients handle multiple discovered servers, server selection, and switching is implementation-defined.
 
 **Note:** After this point, Sendspin works independently of how the connection was established. The Sendspin client is always the consumer of data like audio or metadata, regardless of who initiated the connection.
 
 While custom connection methods are possible for specialized use cases (like remotely accessible web-browsers, mobile apps), most clients should use one of the two standardized methods above if possible.
 
+## Encryption
+
+All Sendspin connections use end-to-end encryption based on the [Noise Protocol Framework](https://noiseprotocol.org/noise.html). Encryption is mandatory for all connections established through the standard discovery mechanisms described in [Establishing a Connection](#establishing-a-connection).
```
maximmaxim345, line 106: Mandatory encryption by default makes sense long-term but how should we handle the transition period where clients aren't updated yet?

With #77 i solved that by allowing unencrypted devices as a advanced option (that we could leave enabled while clients are updated).

README.md
```diff
@@ -280,32 +422,49 @@ Players that can output audio should have the role `player`.
...
 - `name`: string - friendly name of the server
-- `version`: integer (must be `1`) - version of the core message format that the server implements (independent of role versions)
-- `active_roles`: string[] - versioned roles that are active for this client (e.g., `player@v1`, `controller@v1`)
-- `connection_reason`: 'discovery' | 'playback' - only used for [server-initiated connections](#multiple-servers)
+- `connection_reason?`: 'discovery' | 'pairing' | 'playback' | 'management' - only used for [server-initiated connections](#multiple-servers).
   - `discovery` - server is connecting for general availability (e.g., initial discovery, reconnection after connection loss)
+  - `pairing` - server is performing a [pairing](#pairing) handshake
   - `playback` - server needs client for active or upcoming playback
+  - `management` - server is opening a dedicated [management](#management) session - see that section for preconditions and constraints
+- `active_roles?`: string[] - versioned roles that are active for this client (e.g., `player@v1`, `controller@v1`). Required when `connection_reason` is `'playback'`; absent otherwise.
```
maximmaxim345: Before this PR a discovery connection was a full connection. User presses play on the speaker and music just starts. The `connection_reason` field was only there to not kick out other servers that may already play music.

Now discovery has no `active_roles` so the server has to reconnect with `'playback'`. That means a new Noise handshake and the clock-sync filter resets. Playback starts delayed and out of sync until it converges again.

Also L441 says `active_roles` is absent on discovery but L459 says servers always activate the preferred role.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+## Management
+
+This section covers ownership establishment and the management commands an `owner`-trust server may issue.
+
+Management commands are scoped to a **management session**: a connection whose [`connection_reason`](#server--client-serverhello) is `'management'`. The client validates at connection setup that the recorded `trust_level` for the server is `'owner'`; if not, it closes the connection with [`client/goodbye`](#client--server-clientgoodbye) reason `'unauthorized'`. A management session carries no streams; servers MUST NOT send `stream/*` messages on it, and clients SHOULD ignore them if received. If a `management/*` message arrives on a non-management connection, the client replies with [`management/result`](#client--server-managementresult) `permission_denied`.
+
+All `management/*` requests are answered by a single [`management/result`](#client--server-managementresult) message. At most one management request may be in flight per connection; in-order WebSocket delivery makes the reply unambiguous.
+
+### Ownership Claim
```
maximmaxim345: I think no vendor has incentive to defer the claim after the device is paired. Everyone will auto-claim immediately for better UX in their own app. So ownership is decided by install order, not by the user.

User pairs their speaker to LOXONE Audioserver first, that auto-claims as the owner. Later they pair Music Assistant, but that is stuck at the user trust level. Now in Music Assistant you can play music but can't rename the speaker or clear a lockout. The user sees "Music Assistant on this speaker is worse than on my other Sendspin one" and the reason is which app happened to be opened first months ago.

What's the actual benefit of splitting user and owner here? IMO any paired server should be able to do anything.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+- `reason`: string - one of:
+  - `attempt_timeout` (client) - the pairing attempt did not complete within the attempt timeout after [`client/pair-init`](#client--server-clientpair-init) was sent; see [Pairing window](#pairing-window)
+  - `concurrent_attempt` (client) - another pairing attempt is already in progress with this client
+  - `locked_out` (client) - the client is in [terminal lockout](#pin-pairing-lockout) for the selected pairing method
+  - `method_not_supported` (client) - the server's combination of `connection_reason` and `selected_pair_method` is not permitted for the matched PSK, or `selected_pair_method` names a method the client did not list in [`supported_pair_methods`](#client--server-clienthello)
+  - `pin_mismatch` (client or server) - PAKE key-confirmation failed, or (in dynamic PIN pairing) the commitment opening or PIN binding check failed
+  - `storage_exhausted` (client) - client cannot persist a new pairing record and has no fallback policy
+  - `user_cancelled` (client) - operator aborted the pairing through a local UI
+
+## Management
```
maximmaxim345, line 909: The management section adds 13 commands plus result handling and persistence. The rest of the spec is optimized to keep client implementations as minimal as possible (requires the server to support everything, opt-in roles, player chooses what audio format it can support).

Only `set-name` is something a normal user would ever touch. Most of the rest are admin/installer use cases I'm struggling to picture an end-user reaching for. I mean things like `list-records`, `add-record`, `remove-record`, `list-pairing-psks` are really advanced for an average users.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+This section covers ownership establishment and the management commands an `owner`-trust server may issue.
+
+Management commands are scoped to a **management session**: a connection whose [`connection_reason`](#server--client-serverhello) is `'management'`. The client validates at connection setup that the recorded `trust_level` for the server is `'owner'`; if not, it closes the connection with [`client/goodbye`](#client--server-clientgoodbye) reason `'unauthorized'`. A management session carries no streams; servers MUST NOT send `stream/*` messages on it, and clients SHOULD ignore them if received. If a `management/*` message arrives on a non-management connection, the client replies with [`management/result`](#client--server-managementresult) `permission_denied`.
```
maximmaxim345: Just renaming the speaker stops the music. User clicks rename -> server opens a management connection -> priority displaces playback -> music stops. Then another reconnect to come back to playback with the same clock-filter reset.

Just checking if this is intentional.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement either or both PIN methods.
+
+### Methods
+
+1. **Pairing PSK** - pairing authenticated by a [Sendspin Pairing PSK](#definitions); no PAKE round, no PIN. See [Pairing PSK Flow](#pairing-psk-flow).
+2. **Dynamic PIN** - pairing with a per-session [Sendspin Pairing PIN](#definitions); the client derives the PIN from a commit-and-reveal binding to the Noise handshake and emits it via an out-channel (display, speaker, etc.) for the operator to enter into the server. See [Dynamic PIN Pairing Flow](#dynamic-pin-pairing-flow).
+3. **Static PIN** - pairing with a fixed [Sendspin Pairing PIN](#definitions). Appropriate for devices with no out-channel; vulnerable to MITM if the PIN is disclosed. See [Static PIN Pairing Flow](#static-pin-pairing-flow).
```
maximmaxim345, line 648: Static PIN assumes the user can look up a fixed PIN to type in. Realistically that means a sticker printed on the device, which forces manufacturers to decide on Sendspin support before they ship.

For existing devices adopting Sendspin via firmware update there's no sticker.

README.md
```diff
@@ -424,19 +583,491 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+- **Opening the window.** An operator gesture on the client opens the window: a physical button press, a reset-pinhole press, a button combo, a specific power-cycle pattern, a shake or motion gesture, or any equivalent implementation-defined action.
+- **Window lifetime.** From window opening until [`client/pair-init`](#client--server-clientpair-init) is sent. Recommended 5 minutes. On expiry, the window closes silently. A subsequent attempt requires a fresh gesture.
+- **Signal to the server.** The client sends [`client/pair-init`](#client--server-clientpair-init) once the window is open and [`server/hello`](#server--client-serverhello) has arrived. The server must not send [`server/pair-auth`](#server--client-serverpair-auth) until it has received `client/pair-init`.
+
+### PAKE
+
+The PIN pairing flows use **CPACE-X25519-SHA512** as the PAKE construction, defined in [draft-irtf-cfrg-cpace](https://datatracker.ietf.org/doc/draft-irtf-cfrg-cpace/). The protocol runs in initiator-responder mode with explicit Mutual Confirmation Flow (MCF). The server takes role `A` (initiator); the client takes role `B` (responder).
```
maximmaxim345, line 798: How demanding is this?
Is a plain ESP32 (or maybe even an ESP8266) with their CPU/RAM/flash limitations enough?

README.md
arturpragacz: A problem with that approach is downgrade attacks. We should not sanction unencrypted option in the spec, instead we keep it strict and let the server be knowingly non-compliant during the transition period, accepting unencrypted connections until a fixed cutoff.

README.md
arturpragacz: Static PIN is meant to be only a backup option when others are not feasible.

README.md
arturpragacz: I believe it should be fine. We should just implement it and see.

README.md
arturpragacz: Honestly it was a little weird to begin with that you could use connection reason `discovery` to do normal playback.

I had here at first a proper multi-connection handling, but it added a little bit of complexity. Consider for example that a dynamic PIN needs a display/speaker and so cannot be active at the same time as playback is, two management connections at the same time are likely a bad idea too, etc.

So to keep it simple I just limited it to a single connection we had before with the idea that we can extend it in a follow-up.

But if we want, we can try tackling it here already.

Although maybe a simpler alternative for now would be to allow clients to dynamically change their `connection_reason`.

README.md
arturpragacz: See https://github.com/Sendspin/spec/pull/84#discussion_r3267398977.

README.md
arturpragacz: Allowing any paired server to do anything would weaken security: every paired server would need to be hardened to the same (highest) privilege level. It would also shrink the protocol's use cases, since clients could only be shared with fully trusted servers.

That said, you raise a good point that users should be able to easily designate the owner server and change it later, I've added provisions for that to the spec.

README.md
arturpragacz: Users wouldn't necessarily be exposed to those raw commands directly, they'd be used by servers in the background. There are a number of use cases, for example:
- Disabling PINs (Matter does this automatically after first successful pairing, we could do the same).
- Unpairing servers that are no longer in use.
- Sharing a client with another server (e.g. by adding a Pairing PSK to the client and passing it along).

The general idea is to keep the protocol simple and flexible, providing straightforward mechanisms and leaving policy decisions like the above to the servers.

maximmaxim345: Adding 13 commands is anything but keeping the protocol simple.

The 4 commands for just the Pairing PSK need to be implemented by every client. If we want to keep the generate-PSK workflow, I think even a single `management/refresh-pairing-token` should be enough. What's the use case of having multiple long-lived pairing tokens and being able to revoke them individually?

Same goes for the records family of commands. Do we need all 4 of them? `management/add-record` is redundant with the existing pairing mechanisms.
We could maybe even reduce it to one or two commands (a `management/release-ownership` and maybe a `management/unpair-other-servers`), but that looks a bit less flexible to me.

Do we need `management/clear-pin-lockout`? Would a user know to search for this in their main server's settings? I for one wouldn't look there for such an option, maybe it's best to just drop it then? Restarting the speaker or using PSKs is still an option to get the speaker paired.

README.md
```diff
@@ -424,19 +588,504 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pin` rather than `static_pin`. `static_pin` is intended only for devices that genuinely cannot emit a per-session value.
+
+### Unpaired Playback
+
+A client MAY admit `'playback'` connections on the Sentinel PSK from servers with no pairing record. The session's [trust level](#definitions) is `'none'`, so [management](#management) operations remain unavailable. The default is the manufacturer's choice; clients that support the [`controller`](#controller-messages) role SHOULD default to disabled. The toggle is exposed at runtime via [`management/set-pairing-config`](#server--client-managementset-pairing-config), and the client's current setting is advertised in [`client/hello`](#client--server-clienthello) as `unpaired_playback.enabled`. Servers must likewise allow their operator to enable or disable initiating unpaired playback; the server's setting is not exchanged on the wire.
```
maximmaxim345: > clients that support the [`controller`](#controller-messages) role SHOULD default to disabled.

I imagine most speakers/players will also have a play/pause button.
So when going with the recommendation of this spec, you still need to pair almost all speakers before the first use.

To solve this we could restrict some roles to paired sessions only. Right now that would just be the `controller@v1` role, but the same gating would also apply to some future roles like `source@v1`.

We [have a mechanism](https://github.com/Sendspin/spec#server--client-serverhello) to allow only some roles to be active. We just need to adjust this assumption to cover this case as well:

> **Note:** Servers will always activate the client's [preferred](#priority-and-activation) version of each role. Checking `active_roles` is only necessary to detect outdated servers or confirm activation of [application-specific roles](#application-specific-roles).

README.md
maximmaxim345: It's called `connection_reason` since it's the reason why the server initially connected.
 
Could we just allow streams and management commands if the server has the required trust level?
That way the field stays a priority signal without becoming a hard gate.

README.md
```diff
@@ -59,24 +64,23 @@ The server discovers available clients through mDNS and connects to each client
...
+When a new server connects, the client must complete the handshake (send [`client/hello`](#client--server-clienthello), receive [`server/hello`](#server--client-serverhello)) before applying admission. Types are ordered above by priority (management highest, discovery lowest). The incoming connection's priority is compared to the current connection's: higher or equal is accepted, lower is rejected. Two exceptions:
 
-2. **Decide which server to keep**:
-   - If the new server's `connection_reason` is `'playback'` → switch to new server
-   - If the new server's `connection_reason` is `'discovery'` and the existing server connected with `'playback'` → keep existing server
-   - If both servers have `connection_reason: 'discovery'`:
-     - Prefer the server matching the stored last played server
-     - If neither matches (or no history), keep the existing server
+- An in-flight **pairing** is not displaced by another incoming **pairing**.
+- An incoming **discovery**, when an existing **discovery** is held, is admitted iff its `server_id` matches the last-played server (and the existing one's does not); otherwise the existing is kept.
```
kahrendt: Is "iff" a typo or are you using the short hand for "if and only if"? If the shorthand, write it out.

README.md
kahrendt: Is there a security risk to just always include the `active_roles` list? Or at least include it for a `discovery` message so the existing behavior is preserved?

README.md
```diff
@@ -424,19 +588,504 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+- `result`: 'ok' | 'already_owned'
+  - `ok` - the requesting server is now the client's owner. The client records this server's `server_id` as owner and reports `has_owner: true` to all subsequent connections
+  - `already_owned` - the client already had an owner when the request was processed (e.g., the server connected with a stale `has_owner: false`, or `has_owner` flipped between [`client/hello`](#client--server-clienthello) and the claim)
+
+### Server → Client: `management/set-name`
+
+Change the client's friendly `name` (the value reported in subsequent [`client/hello`](#client--server-clienthello) connections and via mDNS).
+
+- `name`: string - new friendly name; non-empty, length cap 128 UTF-8 bytes
```
kahrendt: We have a whole shared key setup for devices with low flash storage size, but now we will require 128 bytes dedicated just to the name?

README.md
kahrendt: The number management messages is my biggest hang up at the moment, this is a significant amount of new messages/complexity to the protocol. I think any reduction we can do here is very beneficial.

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+3. **Static PIN** - pairing with a fixed [Sendspin Pairing PIN](#definitions). Appropriate for devices with no out-channel; vulnerable to MITM if the PIN is disclosed. See [Static PIN Pairing Flow](#static-pin-pairing-flow).
+
+Static pairing methods (Pairing PSK, static PIN) do not take over the device's out-channel. Dynamic pairing (dynamic PIN) takes over the out-channel - typically the audio output or display - to emit the per-session PIN, so it cannot run while audio is playing on the same device. A pairing attempt that arrives while another connection is playing is rejected (see [Multiple servers](#multiple-servers)); the operator must stop playback before initiating pairing.
+
+Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pin` rather than `static_pin`. `static_pin` is intended only for devices that genuinely cannot emit a per-session value.
+
+### Unpaired Playback
```
maximmaxim345: Unpaired playback sounds a bit like it's exclusive to just players, but thats not exactly the case.

I think we can still call it `unpaired_playback` since its consistent with the group state, but maybe add a sentence here clarifying that it's not only for players.

README.md
```diff
@@ -280,31 +442,48 @@ Players that can output audio should have the role `player`.
...
 
-Sends current internal clock timestamp (in microseconds) to the server.
-Once received, the server responds with a [`server/time`](#server--client-servertime) message containing timing information to establish clock offsets.
+Declares the server's current purpose on this connection. Sent as an encrypted message (binary frame, message type `0`). May be re-sent any time to change the activity set.
 
-- `client_transmitted`: integer - client's internal clock timestamp in microseconds
+Only after receiving the initial `server/activate` should the client send any other messages (including [`client/time`](#client--server-clienttime) and the initial [`client/state`](#client--server-clientstate) message if the client has roles that require state updates).
 
-### Server → Client: `server/hello`
+- `activities`: ('playback' | 'pairing' | 'management')[] - the set of currently-active purposes on this connection. May be empty. Members are unordered and unique.
```
maximmaxim345, line 450: We should require the server to keep this minimal, otherwise a server can always include playback (or management), and essentially "break" multi server support by stealing the connection.

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+Possible outcomes: `ok`, `permission_denied`, `already_exists`, `invalid`,  `storage_exhausted`.
+
+#### Record mode
+
+When a server completes pairing via any method, the resulting record is created according to that method's `record_mode`, configured via [`management/set-pairing-config`](#server--client-managementset-pairing-config).
+
+`record_mode` is an object: `{ kind, psk_id? }`.
+
+- `kind: 'individual'`: a new stored-pubkey record is created with a freshly generated per-server PSK (32 bytes from a CSPRNG). On storage exhaustion:
+  - **without `psk_id`**: the pairing fails with [`pair/abort`](#client--server-pairabort) reason `storage_exhausted`.
```
maximmaxim345: Should we specify a minimum storage requirement for records? Without one, a client could ship with only one or two slots and users would hit storage exhaustion much earlier than expected. Hard to write good UX for "you must free a slot" without knowing how many slots exist.

README.md
```diff
@@ -280,31 +442,48 @@ Players that can output audio should have the role `player`.
...
 - `artwork@v1_support?`: object - only if `artwork@v1` is listed ([see artwork@v1 support object details](#client--server-clienthello-artworkv1-support-object))
 - `visualizer@v1_support?`: object - only if `visualizer@v1` is listed ([see visualizer@v1 support object details](#client--server-clienthello-visualizerv1-support-object))
+- `supported_pair_methods?`: object[] - pairing methods this client offers, each described by a [pair-method descriptor](#client--server-clienthello-pair-method-descriptor).
+- `unpaired_playback`: object - whether this client currently admits [unpaired playback](#unpaired-playback)
+  - `enabled`: boolean
 
 **Note:** Each role version may have its own support object (e.g., `player@v1_support`, `player@v2_support`). Application-specific roles or role versions follow the same pattern (e.g., `_myapp_display@v1_support`, `player@_experimental_support`).
 
-### Client → Server: `client/time`
+### Server → Client: `server/activate`
```
maximmaxim345, line 444: What happens to streams when activities transition? Two cases:
- `['playback']` → `['playback', 'management']`: playback stays in the set, streams continue.
- `['playback', 'management']` → `['management']` (or `[]`): playback leaves the set, but management sessions don't have streams. So streams have to stop somehow.

Should we require the server to send `stream/end` for active streams before sending a `server/activate` that drops `'playback'`? Otherwise the client has to handle this edge case itself.

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+- **Increment.** The counter for a method increments on each inner-authentication failure observed in that method's flow.
+- **Reset.** The counter for a method resets to zero on a successful pairing finalization for that method.
+- **Terminal lockout.** When a method's counter reaches **10**, the method enters a **terminal lockout** state: the client refuses all pairing attempts for that method indefinitely. Exit requires a deliberate, local operator action (manufacturer-defined), or writing `locked_out: false` for the method via [`management/set-pairing-config`](#server--client-managementset-pairing-config) from an `owner`-trust server; on successful exit the counter resets to zero. The client SHOULD surface the lockout to the operator via a device-local mechanism (LED, on-screen indicator, audible cue). If a server initiates a pairing-mode connection during terminal lockout, the client sends [`pair/abort`](#client--server-pairabort) with reason `locked_out` and closes.
```
maximmaxim345: This seams a bit dangerous if the client doesn't have an owner to reset the lockout (or the user doesn't know where to find this option).

Also not sure if this should be signaled by an LED or audible cue, especially if Sendspin isn't the only protocol implemented on the device. There will potentially be an always glowing red LED on the device until you do a factory reset.

Can we make this silent and reset on reboot? The default action is to restart hardware if something stops working.

I know that a local action must be provided by the client/device, but having to do a full reset is a pretty bad experience. And requiring manufacturers to create a Sendspin-only reset shortcut is also difficult.

README.md
maximmaxim345: I think the message count is fine now.
We could technically remove `management/add-record` since the same could be achieved without it (by temporarily enabling pairing and sharing the PSK), but keeping it is also fine by me.

README.md
```diff
@@ -16,6 +16,11 @@ Sendspin is a multi-room music experience protocol. The goal of the protocol is
...
+- **Sendspin PSK** - a 32-byte pre-shared symmetric secret shared between a (client, server) pair, established during [pairing](#pairing) and mixed into the [Noise](#encryption) handshake state for every subsequent connection. Must be drawn from a CSPRNG or equivalent high-entropy source.
+- **Sendspin Pairing PSK** - a 32-byte symmetric secret used as the PSK in the [Pairing PSK pairing method](#pairing). It is always distributed alongside the client's static public key (`client_id`), which the server needs to verify the client identity. The operator enters it into the server by copying a string or scanning a QR code. Distinct from the per-pair Sendspin PSK that pairing produces. Must be drawn from a CSPRNG or equivalent high-entropy source.
+- **Sendspin Pairing PIN** - an 8-decimal-digit value used in PIN-based [pairing](#pairing) methods. The static-PIN method uses a fixed value, the dynamic-PIN method uses a per-session generated value.
```
balloob: Dynamic pin is 4 digits, static 8 digits.

README.md
```diff
@@ -16,6 +16,11 @@ Sendspin is a multi-room music experience protocol. The goal of the protocol is
...
+- **Sendspin Trust Level** - one of `owner`, `user`, or `none`, expressing the trust the client extends to the server. Ordered `none < user < owner`. `owner` and `user` are recorded client-side per pairing record; `none` is not stored but is the effective trust level on any connection where no pairing record exists for the peer. Servers with `owner` trust may issue [management commands](#management) to the client; servers with `user` trust are restricted to normal playback and control flows; and servers with `none` trust are further restricted to conducting a pairing exchange or, when [unpaired playback](#unpaired-playback) is enabled, normal playback and control flows.
```
balloob: I feel like adding `owner` here is making things complicated from the get go. How will a speaker without a screen ever make this clear? And how would a user know which one is that? we should not include `owner`.

README.md
```diff
@@ -29,7 +34,7 @@ All role names and versions not starting with `_` are reserved for future revisi
 
 Clients list roles in `supported_roles` in priority order (most preferred first). If a client supports multiple versions of a role, all should be listed: `["player@v2", "player@v1"]`.
 
-The server activates one version per role family (e.g., one `player@vN`, one `controller@vN`)—the first match it implements from the client's list. The server reports activated roles in `active_roles`.
+The server activates at most one version per role family (e.g., one `player@vN`, one `controller@vN`) - the first match it implements from the client's list, or none if server policy declines to activate that family. The server reports activated roles in `active_roles`; clients MUST consult it and refrain from sending commands or state for roles that aren't active.
```
balloob, line 35: This change seems semi-related, and could probably be PRd separely to tighten language.

README.md
```diff
@@ -59,24 +64,26 @@ The server discovers available clients through mDNS and connects to each client
...
-#### Multiple Servers
+#### Multiple servers
 
-In environments with multiple Sendspin servers, servers may need to reconnect to clients when starting playback to reclaim them. The [`server/hello`](#server--client-serverhello) message includes a `connection_reason` field indicating whether the server is connecting for general availability (`'discovery'`) or for active/upcoming playback (`'playback'`).
+A client holds at most one admitted connection at a time, classified by the highest-ranked activity in its declared [`activities`](#server--client-serveractivate); from highest to lowest:
 
-Clients can only be connected to one server at a time. Clients must persistently store the `server_id` of the server that most recently had `playback_state: 'playing'` (the "last played server").
+- `'management'`
+- `'playback'`
+- `'pairing'`
```
balloob, line 71: Why do we care about this? 

A connection comes in, it is not encrypted: it's `pairing`.

A connection comes in, it's encrypted: it's everything else.

README.md
balloob: Same with capitalization of headers.

README.md
```diff
@@ -90,22 +97,108 @@ Clients discover the server through mDNS and initiate a WebSocket connection usi
...
 
 **Note:** After this point, Sendspin works independently of how the connection was established. The Sendspin client is always the consumer of data like audio or metadata, regardless of who initiated the connection.
 
 While custom connection methods are possible for specialized use cases (like remotely accessible web-browsers, mobile apps), most clients should use one of the two standardized methods above if possible.
 
+## Encryption
+
+All Sendspin connections use end-to-end encryption based on the [Noise Protocol Framework](https://noiseprotocol.org/noise.html). Encryption is mandatory for all connections established through the standard discovery mechanisms described in [Establishing a Connection](#establishing-a-connection).
+
+Specialized deployments where the connection is tunneled through a separately authenticated and encrypted channel may expose a non-standard endpoint that omits the Noise handshake; such endpoints must not be advertised via mDNS, and the operator is responsible for ensuring the tunnel provides equivalent security guarantees.
```
balloob: I think that we should drop this from the spec. If these kind of things are done, they are not spec compliant and it's up to the project to make sure it works.

README.md
```diff
@@ -90,22 +97,108 @@ Clients discover the server through mDNS and initiate a WebSocket connection usi
...
+
+### Pattern
+
+Sendspin uses the `KKpsk2` Noise pattern. Both static keys are pre-known to both parties (the `client_id` of the client and the `server_id` of the server are the static public keys), and a [Pre-Shared Key](#pre-shared-key) is mixed in at the end of the handshake's second message.
+
+The **server is the Noise initiator**, the **client is the Noise responder**, regardless of which side initiated the WebSocket connection.
+
+**Security properties.** Forward secrecy is provided by the ephemeral-key DH in each handshake: compromise of static keys or the PSK does not retroactively decrypt prior sessions. Replay protection is provided by Noise's per-direction transport counter; a repeated or out-of-order ciphertext fails AEAD decryption and aborts the connection.
+
+### Cipher Suites
```
balloob, line 116: Some headers have been adjusted to lowercase the 2nd word, here we capitalize 2nd word.

README.md
```diff
@@ -90,22 +97,108 @@ Clients discover the server through mDNS and initiate a WebSocket connection usi
...
+- **Shared-PSK model**: PSKs are persisted without an associated `server_id`; the `server_id` from [`server/init`](#server--client-serverinit) is accepted at face value. Convenient for storage-constrained clients, but with weaker security properties - multiple servers may share the same PSK.
+
+### Prologue
+
+The prologue mixed into the Noise handshake state on both sides is the concatenation of the exact bytes of [`client/init`](#client--server-clientinit) followed by the exact bytes of [`server/init`](#server--client-serverinit), as transmitted on the wire (the JSON-encoded UTF-8 message body, without the WebSocket framing). This binds the cleartext init exchange to the handshake; tampering causes the handshake to fail.
+
+### Failure Handling
+
+Any handshake-phase failure - malformed cleartext message, unsupported `version`, unknown `suite`, handshake timeout, `psk_id` lookup miss, Noise AEAD failure, or AEAD failure once in transport mode - closes the WebSocket silently. Implementations SHOULD apply a timeout (e.g., 30 seconds) for each side to receive the next expected message during the prologue and Noise-handshake phases.
```
balloob: Why would unsupported `version` cause the socket to be closed silently? That means that a pairing failure is not known until 30s later and we don't know what exactly went wrong?

README.md
```diff
@@ -90,22 +97,108 @@ Clients discover the server through mDNS and initiate a WebSocket connection usi
...
 ## Communication
 
-Once the connection is established, Client and Server are going to talk.
+Once the WebSocket connection is established, Client and Server perform an initial handshake before exchanging application data:
 
-The first message must always be a `client/hello` message from the client to the server.
-Once the server receives this message, it responds with a `server/hello` message. Before this handshake is complete, no other messages should be sent.
+1. Client → Server: [`client/init`](#client--server-clientinit) (cleartext)
+2. Server → Client: [`server/init`](#server--client-serverinit) (cleartext)
+3. Server → Client: [`noise/handshake`](#client--server-noisehandshake) - Noise message 1 (cleartext)
```
balloob, line 186: We can merge these 2 messages. Why make them separate?

README.md
```diff
@@ -137,14 +230,16 @@ Message format example:
 }
 ```
 
-WebSocket binary messages are used to send audio chunks, media art, and visualization data. The first byte is a uint8 representing the message type.
+WebSocket binary messages are used to send JSON payloads, audio chunks, media art, and visualization data. Each binary message is a Noise transport ciphertext; after AEAD decryption, the first byte is a uint8 representing the message type.
```
balloob:
```suggestion
WebSocket binary messages are used to send encrypted JSON payloads, audio chunks, media art, and visualization data. Each binary message is a Noise transport ciphertext; after AEAD decryption, the first byte is a uint8 representing the message type.
```

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+### Unpaired Playback
+
+A client MAY admit `'playback'` connections on the Sentinel PSK from servers with no pairing record. The session's [trust level](#definitions) is `'none'`, so [management](#management) operations remain unavailable. Servers SHOULD consider their role-activation policy on such sessions in light of the MITM exposure described below - in particular, the [`controller`](#controller-messages) role lets the client issue commands that affect its entire group and is a reasonable candidate to omit on `'none'`-trust sessions. The default is the manufacturer's choice. The toggle is exposed at runtime via [`management/set-pairing-config`](#server--client-managementset-pairing-config), and the client's current setting is advertised in [`client/hello`](#client--server-clienthello) as `unpaired_playback.enabled`. Servers must likewise allow their operator to enable or disable initiating unpaired playback, with the current setting advertised in [`server/hello`](#server--client-serverhello).
```
balloob: Things like `unpaired_playback` and `management/set-pairing-config` are out of scope for an initial version of encryption. Let's keep it simple and we just allow this.

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+**Security.** Unpaired playback connections are vulnerable to **man-in-the-middle attacks**. The Sentinel PSK is a published constant, and the peer's static key is learned from mDNS, which is unauthenticated; an attacker on the local network may therefore impersonate either side. The Noise handshake still provides confidentiality and replay protection for the session itself, but offers no assurance about which peer it was established with.
```
balloob: I don't think the peer's static key is learned from mDNS.

From line 116:

> the `client_id` of the client and the `server_id` of the server are the static public keys

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+sequenceDiagram
+    participant Client
+    participant Server
+
+    Note over Client,Server: Noise handshake completes with Sentinel PSK
+
+    Server->>Client: server/hello (name)
+    Client->>Server: client/hello (supported_pair_methods)
+    Note over Server: Operator picks dynamic PIN
+    Server->>Client: server/activate (activities=['pairing'], selected_pair_method=dynamic_pin)
```
balloob: why is this `server/activate` and not `server/pair` ? And if we receive that, track that current client is in pairing mode.

One thing we see by re-using `server/activate` for multiple purposes, is that we are including fields for one of the purpopses (ie `selected_pair_method`)

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+## Management
+
+This section covers ownership establishment and the management commands an `owner`-trust server may issue.
+
+Management commands are scoped to connections with `'management'` in their [`activities`](#server--client-serveractivate). When the server adds `'management'` to the activity set, the client validates that the recorded `trust_level` for the server is `'owner'`; if not, it closes the connection with [`client/goodbye`](#client--server-clientgoodbye) reason `'unauthorized'`. If a `management/*` message arrives on a connection without `'management'` in activities, the client replies with [`management/result`](#client--server-managementresult) `permission_denied`.
+
+All `management/*` requests are answered by a single [`management/result`](#client--server-managementresult) message. At most one management request may be in flight per connection; in-order WebSocket delivery makes the reply unambiguous.
+
+### Ownership Claim
```
balloob: let's drop this from the initial encryption proposal.

README.md
```diff
@@ -424,19 +603,464 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+#### Client → Server: `client/claim-ownership-result`
+
+Response to [`server/claim-ownership`](#server--client-serverclaim-ownership).
+
+- `result`: 'ok' | 'already_owned'
+  - `ok` - the requesting server is now the client's owner. The client records this server's `server_id` as owner and reports `has_owner: true` to all subsequent connections
+  - `already_owned` - the client already had an owner when the request was processed (e.g., the server connected with a stale `has_owner: false`, or `has_owner` flipped between [`client/hello`](#client--server-clienthello) and the claim)
+
+### Records
```
balloob, line 917: Why would we need this? Can we just start by storing the last 5 pairings?

README.md
arturpragacz: I added a way for clients to inform how much storage they have left.

I'd rather not be too restrictive towards small clients, that's why we have fallback to shared records in the first place.

README.md
arturpragacz: Why would client need to worry about it? Presenting that information is server's job.

README.md
arturpragacz: We need to distinguish it for multi server arbitration purposes.

README.md
arturpragacz: They have very different semantic meaning, why would we merge them?

README.md
arturpragacz: I strongly disagree. I'm not a fan of the fact that we have unpaired access to begin with, but if it's here, it definitely needs to be possible to disable.

README.md
arturpragacz: I softened the indicator requirements here.

README.md
marcelveldt: What about we keep it simple and just define in the spec that a client should at least have 5 slots available - higher performance clients can have more slots. Then its just FIFO. Pair a new connection, drops the oldest.

README.md
marcelveldt: same comment here - I think this makes it way too complex, especially for a v1 version

README.md
```diff
@@ -280,31 +440,54 @@ Players that can output audio should have the role `player`.
...
-Once received, the server responds with a [`server/time`](#server--client-servertime) message containing timing information to establish clock offsets.
+Declares the server's current purpose on this connection. Sent as an encrypted message (binary frame, message type `0`). May be re-sent any time to change the activity set.
 
-- `client_transmitted`: integer - client's internal clock timestamp in microseconds
+Only after receiving the initial `server/activate` should the client send any other messages (including [`client/time`](#client--server-clienttime) and the initial [`client/state`](#client--server-clientstate) message if the client has roles that require state updates).
 
-### Server → Client: `server/hello`
+- `activities`: ('playback' | 'pairing' | 'management')[] - the set of currently-active purposes on this connection. May be empty. Members are unordered and unique.
+- `active_roles?`: string[] - versioned roles that are active for this client (e.g., `player@v1`, `controller@v1`), must be empty on connections not capable of playback. Required on the first `server/activate`; persists across subsequent `server/activate` messages that omit it.
```
maximmaxim345: "capable of playback" is never defined. When I read this I understood that this either means `playback` is in `activities` (contradicts below with the 'A playback-capable connection MAY carry a non-empty ...' note) or "the client can output audio" (which contradicts with controller only devices).

README.md
```diff
@@ -280,31 +440,54 @@ Players that can output audio should have the role `player`.
...
-- `version`: integer (must be `1`) - version of the core message format that the server implements (independent of role versions)
-- `active_roles`: string[] - versioned roles that are active for this client (e.g., `player@v1`, `controller@v1`)
-- `connection_reason`: 'discovery' | 'playback' - only used for [server-initiated connections](#multiple-servers)
-  - `discovery` - server is connecting for general availability (e.g., initial discovery, reconnection after connection loss)
-  - `playback` - server needs client for active or upcoming playback
+¹ `['playback']` on the Sentinel PSK is only allowed when the client has [unpaired access](#unpaired-access) enabled.
+
+`selected_pair_method` must additionally match the `method` field of one of the [pair-method descriptors](#client--server-clienthello-pair-method-descriptor) the client listed in [`supported_pair_methods`](#client--server-clienthello).
+
+A playback-capable connection MAY carry a non-empty `active_roles` even when `'playback'` is not currently in `activities`. On the Sentinel PSK, a non-empty `active_roles` requires [unpaired access](#unpaired-access).
```
maximmaxim345: This is what enables an idle connection to: "show what will be resumed and pressing play will start the music". 
But this is left entirely to the server implementation.
I think it would be best to add a SHOULD to keep display roles (`metadata@v1`, `artwork@v1`, ..., or however how or what we define as display roles).

Otherwise the natural reading "playback activity left, so tear down all streams" kills the idle artwork.
It should be a client/implementation decision whether to show artwork when not playing.

README.md
```diff
@@ -425,19 +608,493 @@ Contains delta updates with only the changed fields. The client should merge the
...
+    Client->>Server: client/pair-finalize (long_term_psk)
+    Server->>Client: server/pair-finalize
+    Note over Client,Server: Both sides persist the pairing record. Server re-handshakes to long_term_psk.
+```
+
+**Binding values.** The dynamic PIN flow introduces three values across two messages that bind the PIN to the underlying Noise handshake:
+
+- `nonce_A` - 32 bytes drawn from a CSPRNG by the server, sent in [`server/pair-init`](#server--client-serverpair-init), base64url-encoded (43 chars).
+- `nonce_B` - 32 bytes drawn from a CSPRNG by the client, kept private until [`client/pair-confirm`](#client--server-clientpair-confirm) reveals it (base64url-encoded, 43 chars).
+- `commit_B` - `SHA-256(nonce_B)`, sent by the client in [`client/pair-init`](#client--server-clientpair-init) before any value from the server is known (32 bytes base64url-encoded, 43 chars). Locks the client's contribution to the PIN derivation.
```
maximmaxim345, line 724: Why is `L` negotiated per attempt?

The only thing I can think of is that this could enable growing pin lengths (a shorter first attempt, longer on retry as a compromise for an too-short PIN), but this is just a guess I have. 
As written this sounds like a complication searching for a use case.

README.md
```diff
@@ -425,19 +608,493 @@ Contains delta updates with only the changed fields. The client should merge the
...
+
+**Device-presence verification.** When the matched PSK is a long-term [Sendspin PSK](#definitions), the dynamic-PIN sequence runs through the PAKE round and commitment reveal, but the two `pair-finalize` messages are omitted: no new long-term PSK is established. A failed check has no effect on the existing pairing record and does not increment the PIN-pairing [failure counter](#pin-pairing-lockout). After verifying [`client/pair-confirm`](#client--server-clientpair-confirm), the server sends a fresh [`server/activate`](#server--client-serveractivate) to [leave pairing](#entering-and-leaving-pairing), reactivating the client's roles. The purpose is to confirm that the device holding the long-term PSK is the same physical device the operator is currently observing - useful on top of static pairing methods, which establish cryptographic identity but do not bind it to a specific physical device.
```
maximmaxim345: Why do we need this?
Will a user/operator ever realistically use this?

README.md
arturpragacz: The next sentence already says they're all encrypted.

README.md
arturpragacz: This is fundamentally up to the server implementation, as it depends on what the server does exactly when it stops playing. There is no guarantee that there is any sort of queue that is stopped, from which to take artwork. I don't think we should mandate any specific behaviour here.

maximmaxim345: Makes sense, I think it's also a lot clearer now with the "Playback-capable" explanation you added (so it's no longer a "hidden" feature of the protocol).

README.md
```diff
@@ -280,31 +439,54 @@ Players that can output audio should have the role `player`.
...
+
+**Note:** Servers normally activate the client's [preferred](#priority-and-activation) version of each role, but MAY omit a role at their discretion (e.g., based on trust level, deployment context, or operator policy). Checking `active_roles` is therefore required to determine what the client may actually use on this session.
 
-**Note:** Servers will always activate the client's [preferred](#priority-and-activation) version of each role. Checking `active_roles` is only necessary to detect outdated servers or confirm activation of [application-specific roles](#application-specific-roles).
+**Note:** When a `server/activate` removes a role from `active_roles`, the server first sends [`stream/end`](#server--client-streamend) for any stream that role still has open, so the client never holds a live stream for an inactive role.
```
kahrendt: `stream/end` only applies to binary message roles. Should we specify roles that just send JSON; e.g., color and metadata, should clear all fields before removing those roles?

README.md
arturpragacz: It is actually a consequence of the simplification that we made - we currently allow exactly one attempt per pairing session. I adjusted the wording.

README.md
arturpragacz: I think it's useful, but it's also more about not artificially blocking it, all the functionality is there already.

I now simplified it even more.

README.md
marcelveldt: we'll do a textual pass afterwards

kahrendt (approved):

maximmaxim345 (approved):
🎉

PR #86, opened 2026-05-21
Add `visualizer@v1` role

maximmaxim345:
Adds the visualizer role, based on [the previous `visualizer@_draft_r1` proposal](#28).

This alternative version aims to resolve a couple of gaps that were open in the previous PR:

## One frame per binary message
Batching multiple frames of mixed types into one WebSocket message, as in the previous proposal, forces the server to either delay frames waiting for siblings (hurting low-latency playback) or send tiny batches anyway. It also makes ordering awkward across batches and pushes sort work onto the client. Per-type messages keep ordering trivial. Also more consistent with how other roles structure their binary messages.

The only concern with this approach is the increased number of messages. We could alternatively batch multiple message types with the same timestamp together.

## Downbeat flag on `beat`
Lets clients drive bar-aware effects. `stream/start` advertises `tracks_downbeats` so clients know whether to trust the bit. Accurate beat detection is hard and often relies on offline analysis, so servers without it omit the `beat` type entirely. Even when supported, it may be unavailable for some content (live streams, sparse non-percussive material).

## Top-level `rate_max`
Now bounds all periodic types, not just `spectrum`. `beat` and `peak` are event-driven and unthrottled.

## Scaling
Pins down what was previously hand-waved as "perceptual weighting" so implementations agree on the numbers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -746,7 +746,6 @@ The `visualizer` object in [`stream/request-format`](#client--server-streamreque
 - `visualizer`: object
   - `types?`: string[] - new set of visualization data types
   - `rate_max?`: integer - new periodic frames-per-second cap
-  - `buffer_capacity?`: integer - new buffer capacity in bytes
   - `spectrum?`: object - new spectrum configuration ([see spectrum object details](#client--server-clienthello-visualizerv1-support-object))
 
 All fields are optional; omitted fields keep their current value.
--- a/README.md
+++ b/README.md
@@ -791,7 +791,7 @@ Tracks the dominant FFT bin, which is not always the fundamental: strong harmoni
 
 - 2*n bytes: `uint16[n]` bins from low to high frequency. `n` = `n_disp_bins` in [`stream/start`](#server--client-streamstart-visualizer-object)
 
-Magnitude per display bin. Servers may impose an implementation-defined upper bound on `n_disp_bins` to keep per-frame size sensible.
+Magnitude per display bin (see scaling above). Servers may impose an implementation-defined upper bound on `n_disp_bins` to keep per-frame size sensible.
 
 #### `peak` — message type `20`
 
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Message format example:
 }
 ```
 
-WebSocket binary messages are used to send audio chunks, media art, and visualization data. The first byte is a uint8 representing the message type.
+WebSocket binary messages are used to send audio chunks, media art, and visualization data. The first byte is a uint8 representing the message type. Throughout this specification, bit 0 refers to the least significant bit.
 
 ### Binary Message ID Structure
 
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ sequenceDiagram
             Server->>Client: binary Types 8-11 (artwork channels 0-3)
         end
         alt Visualizer role
-            Server->>Client: binary Type 16 (visualization data)
+            Server->>Client: binary Types 16-20 (loudness, beat, f_peak, spectrum, peak)
         end
     end
 
@@ -395,10 +395,11 @@ Instructs clients to clear buffers without ending the stream. Used for seek oper
 
 ### Client → Server: `stream/request-format`
 
-Request different stream format (upgrade or downgrade). Available for clients with the `player` or `artwork` role.
+Request different stream format (upgrade or downgrade). Available for clients with the `player`, `artwork`, or `visualizer` role.
 
 - `player?`: object - only for clients with the `player` role ([see player object details](#client--server-streamrequest-format-player-object))
 - `artwork?`: object - only for clients with the `artwork` role ([see artwork object details](#client--server-streamrequest-format-artwork-object))
+- `visualizer?`: object - only for clients with the `visualizer` role ([see visualizer object details](#client--server-streamrequest-format-visualizer-object))
 
 [Application-specific roles](#application-specific-roles) may also include objects in this message (keys starting with `_`).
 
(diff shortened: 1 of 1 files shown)


jhollowe:
@Aircoookie this looks to be the successor to #28. Does this include all the information you would like available?

README.md
```diff
@@ -704,36 +705,107 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
 **Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty binary message (only the message type byte and timestamp, with no image data) for that channel.
 
 ## Visualizer messages
-This section describes messages specific to clients with the `visualizer` role, which create visual representations of the audio being played. Visualizer clients receive audio analysis data like FFT information that corresponds to the current audio timeline.
+This section describes messages specific to clients with the `visualizer` role, which create visual representations of the audio being played. Visualizer clients receive audio analysis data computed from the audio currently playing in the group.
+
+Each visualizer binary message carries exactly one frame. The server emits messages in non-decreasing timestamp order so clients can process them in arrival order. Types the server cannot stream for the current source are silently omitted from the set echoed in [`stream/start`](#server--client-streamstart-visualizer-object). `beat` and `peak` are event-driven and not throttled by `rate_max`; all other types are periodic.
```
maximmaxim345, line 710: For the Music Assistant server implementation the non-decreasing timestamp order is the most difficult part of this PR.

Beats are computed in the background and aren't immediately available, there is a conflict between:
- maximizing the buffer size of visualization data
- and sending the analyzed beats to the client as soon as they are available

So, during track transitions we might not have beats data.

My workaround for this is:
- keeping the whole visualizer buffer limited at only 3s.
- and re-sending `stream/start` with and without `beats` so clients know when beats are available and when not.

README.md
```diff
@@ -704,36 +705,107 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+
+All fields are optional; omitted fields keep their current value.
+
+Response: [`stream/start`](#server--client-streamstart) with the new visualizer configuration.
 
 ### Server → Client: `stream/clear` visualizer
 
 When [`stream/clear`](#server--client-streamclear) includes the visualizer role, clients should clear all buffered visualization data and continue with data received after this message.
 
 ### Server → Client: Visualization Data (Binary)
```
maximmaxim345, line 759: Another potential concern is the amount of messages we are sending.
I don't think the overhead of WebSocket messages is too large though so this just needs testing on more low powered hardware.

There are two reasons why messages are completely split now:
- Consistency with other roles, all other roles already have one message per datum
- Difficulty of defining batching behavior. Requiring batching of multiple messages is difficult since it's always a compromise between latency and message count. But leaving batching open to the server would cause most implementations to never use them, defeating the whole purpose.

README.md
maximmaxim345: I have no better solution though, two ideas are:
1. Allowing decreasing timestamps: makes parsing more difficult on the resource constrained client side
2. Ensuring timestamp is non-decreasing per type instead of globally: Client implementations most likely want to use the different types together. So we are back at the same issues as idea 1.

So the best path is probably to keep the spec as is and just optimize Music Assistant to better pre-compute beats so we can increate the buffer again.

README.md
maximmaxim345: In case this is really a problem, we could also later release a `visualizer@v2` if the overhead turns out to be bigger than expected.

This just needs to be tested (with encryption) on a ESP8266 or similar.

README.md
```diff
@@ -704,36 +705,107 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+
+Magnitude per display bin. Servers may impose an implementation-defined upper bound on `n_disp_bins` to keep per-frame size sensible.
+
+#### `peak` — message type `20`
+
+- 1 byte: `uint8` strength
+
+Energy onset event. Fires on any transient (drum hits, cymbal crashes, attacks), independent of musical timing. `strength` 0-255 lets clients scale flash intensity.
+
+#### `pitch` — message type `21`
```
maximmaxim345: I have a couple concerns with pitch that came up while implementing this into sendspin-cli.

- First of all, the pitch given by `aiosendspin` isn't too precise and useful. (disabled in Music Assistant for this reason). But thats more of an implementation issue.
- Secondly, how long is a pitch supposed to be valid? If the server stops emitting when there's nothing tonal, the last value just sticks until the next track or so. One could interpret "ignore below your own threshold" as "clear when confidence is below your threshold or 0", but thats not defined in the Specification.
- And then the confidence scale itself: every client picking its own threshold and no defined meaning of the threshold makes behavior between server and client implementations inconsistent.

But if there is no reliable way to get a single useful pitch value, we could also just consider removing `pitch` from the specification.

maximmaxim345: Dropped the `pitch` type for now, we can later add it again if we are more confident that it really should be part of the spec.

README.md
```diff
@@ -704,36 +705,100 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+    - `f_max`: integer - highest frequency in Hz
+
+### Client → Server: `stream/request-format` visualizer object
+
+The `visualizer` object in [`stream/request-format`](#client--server-streamrequest-format) has this structure:
+
+- `visualizer`: object
+  - `types?`: string[] - new set of visualization data types
+  - `rate_max?`: integer - new periodic frames-per-second cap
+  - `buffer_capacity?`: integer - new buffer capacity in bytes
```
kahrendt: This seems unusual. The player role doesn't allow changing the buffer capacity outside of the hello message. What use case would this have for mid-stream?

README.md
```diff
@@ -704,36 +705,100 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+
+#### `loudness` — message type `16`
+
+- 2 bytes: `uint16` value
+
+Overall A-weighted loudness in dB (see scaling above).
+
+#### `beat` — message type `17`
+
+- 1 byte: `uint8` flags. Bit 0 = downbeat (bar start). Bits 1-7 reserved, must be zero by the server, ignored by the client
```
kahrendt, line 779: We should maybe add a general note to the spec that bit 0 is the least significant bit for clarification (this applies to the binary message type)

maximmaxim345: Done, added a note in the "Communication" section.

README.md
maximmaxim345: I kept it in so clients could reuse the same object in both `client/hello` and `stream/request-format`.
But yea, I can see a lot of new edge cases when the client changes the buffer size mid-stream.
Removed it.

README.md
```diff
@@ -704,36 +705,99 @@ The timestamp indicates when this artwork should be displayed. Clients must tran
...
+- 2 bytes: `uint16` freq - dominant frequency in Hz (0 = no peak detected, amp must also be 0)
+- 2 bytes: `uint16` amp - amplitude (see scaling above)
+
+Tracks the dominant FFT bin, which is not always the fundamental: strong harmonics can dominate, so do not treat `f_peak` as the musical note being played.
+
+#### `spectrum` — message type `19`
+
+- 2*n bytes: `uint16[n]` bins from low to high frequency. `n` = `n_disp_bins` in [`stream/start`](#server--client-streamstart-visualizer-object)
+
+Magnitude per display bin. Servers may impose an implementation-defined upper bound on `n_disp_bins` to keep per-frame size sensible.
```
kahrendt: The other two types have "(see scaling above)", it would be nice if this had it as well for consistency.,

kahrendt (approved):
1 small nit but overall looks good!

README.md
maximmaxim345: Done.

Issue #91, opened 2026-05-27
Goodbye and operational state: default reason, external_source API, delta updates


balloob:
_This issue comes out of a cross-SDK conformance audit comparing every Sendspin client/server implementation (`aiosendspin`, `sendspin-cli`, `sendspin-cpp`, `sendspin-dotnet`, `sendspin-go`, `sendspin-js`, `sendspin-jvm`, `sendspin-rs`, `SendspinKit`) against the spec. **`sendspin-cpp` is treated as the reference implementation** throughout — it most closely matches the spec's language and is the only SDK that runs on the constrained-embedded target the spec was originally written for._

Most SDKs implement the message envelopes correctly but only one
(`SendspinKit`) implements `external_source` as a client API, and
several SDKs default to the wrong goodbye reason in common situations.
For `external_source`, **`SendspinKit` is the reference** —
`enterExternalSource()` / `exitExternalSource()` route through
`transitionOperationalState(to:)` with rollback on
server-notification failure. For goodbye-reason plumbing,
`sendspin-cpp` is the reference (`SendspinGoodbyeReason` enum,
auto-`another_server` from `ConnectionManager`).

### 1. Pick a default goodbye reason and make it consistent

> When a client disconnects without specifying a reason in its SDK API,
> the SDK SHOULD send `reason: 'restart'`, matching the server's
> no-goodbye assumption.

Today defaults are unprincipled: `sendspin-js` → `shutdown`,
`sendspin-dotnet` → `user_request`, `SendspinKit` → `shutdown`. All
three suppress server-side auto-reconnect for a user who simply
closes a laptop lid — the opposite of what the spec implies.

### 2. Add a normative note about the `external_source` SDK API

> Client SDKs SHOULD expose API methods for entering and leaving
> `external_source` state and SHOULD roll back the local state on
> server-notification failure.

SendspinKit's rollback pattern
(`SendspinClient+Commands.swift:27-45`) is the model. Eight of nine
SDKs have no embedder-facing handle today — so a Sonos-style "input
switched to TV" event cannot be relayed to the server, and the
server keeps streaming audio into silence.

### 3. Document or relax delta `client/state` merge behavior

The spec says: "for subsequent updates, only include fields that have
changed; the server will merge these updates into existing state".
**No SDK actually sends delta updates** — everyone sends the full
payload every time. Either:

- add normative test cases (e.g., "if the client previously sent
  `static_delay_ms: 100` and now sends `{volume: 50}`, the server's
[...]

PR #101, opened 2026-06-15
Clarify client/state delta vs full updates

### LINKED ISSUE #98: `client/state` delta rule contradicts "always required" player fields
`client/state` says deltas "only include fields that have changed," but `static_delay_ms`, `required_lead_time_ms`, and `min_buffer_ms` are each marked "always required for players." A delta changing only one field cannot satisfy both rules.

The player-object descriptions should state whether these three may be omitted from deltas when unchanged.

kahrendt:
`client/state` said deltas "only include fields that have changed," but `static_delay_ms`, `required_lead_time_ms`, and `min_buffer_ms` were marked "always required for players." A delta changing one field can't satisfy both rules (#98). Separately, the spec didn't make clear whether a client may send full state instead of a delta, or what the server does with an absent field (#91, item 3).

Changes:

- Drop "always" from the three player fields and scope all field-presence rules to a player's full state.
- State the update model once: the initial message MUST be complete; later messages MAY send only changed fields; the server MUST merge and retain the last value of any absent field; a client MAY resend unchanged fields up to its full state.
- Replace the per-field delta wording with a single "Delta updates" note.
- Use RFC 2119 keywords in the affected passages.

A missing field always means "unchanged" — `client/state` has no null-clearing semantics, so the server never has to tell a delta apart from a full update; it merges either one the same way.

Out of scope: the other two items in #91 (default goodbye reason, `external_source` SDK API).

Closes #98
Refs #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ Client sends state updates to the server. Contains client-level state and role-s
 
 Must be sent immediately after receiving [`server/hello`](#server--client-serverhello), and whenever any state changes thereafter.
 
-For the initial message, include all state fields. For subsequent updates, only include fields that have changed. The server will merge these updates into existing state.
+The initial message MUST include all state fields. In subsequent messages, the client MAY send only the fields that have changed; the server MUST merge each update into existing state, retaining the last value of any field that is absent. A client MAY instead resend unchanged fields, up to its full state.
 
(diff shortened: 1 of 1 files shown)


maximmaxim345 (approved):
LGTM.

Issue #88, opened 2026-05-27
Specify a late-chunk drop threshold for players


balloob:
_This issue had three parts. Two parts are complete after the change to the Playback Synchronization section. See the comment below. Only the late-chunk threshold stays._

## Problem

The section Sync Accuracy gives one sentence about late chunks:

> Audio chunks may arrive with timestamps in the past due to network delays or buffering; clients should drop these late chunks to maintain sync.

The spec gives no threshold. Thus a conformance audit found a large difference between the SDKs. `sendspin-cpp` and `sendspin-cli` use 5 ms. `sendspin-go` and `SendspinKit` use 50 ms. `sendspin-jvm` uses 1 s.

Two clients in one group then have different behavior with the same network conditions. One client drops the chunk and synchronizes again. The other client plays the chunk late.

This is the difference that the accuracy limit of ±1 ms prevents.

## Change

Add this text:

> Clients SHOULD drop any audio chunk whose scheduled play time is more than **5 ms** before the end of currently buffered/playing audio. Clients MAY trim the stale prefix of a chunk whose tail is still in the future.

This threshold agrees with `sendspin-cpp`. It also agrees with the trim function in `sendspin-cli`.

## Related decision

The spec has three related limits: this threshold, the accuracy limit of ±1 ms, and the single correction in "Large errors and startup".

Show how the three limits agree. If they do not agree, they give three different recovery points for one condition.

---

**Source audit doc:** [`docs/stream-sync-correction.md`](https://github.com/Sendspin/conformance/blob/claude/stream-sync-correction-sdks-AWoNC/docs/stream-sync-correction.md)
**Full audit branch:** [`claude/stream-sync-correction-sdks-AWoNC`](https://github.com/Sendspin/conformance/tree/claude/stream-sync-correction-sdks-AWoNC/docs)
**Per-SDK digest:** [sdk-issues digest](https://gisthost.github.io/?ac84f17e71e7eccc48c9f08d5df0a7b4)</body>
</invoke>

balloob:
Re-audited against the current spec and re-scoped the issue to the one remaining item. Record of what changed:

[...]

PR #102, opened 2026-06-15
Simplify client/state status

kahrendt:
Reworks the client state model so a client reports synchronization status rather than errors.

- Removes the `error` state
- Clients only send the initial `client/state` message when time is synchronized
- The server MUST NOT send binary messages (audio, artwork, visualization) to a client until the client reports the synchronized state. Each binary section now rejects messages outside that state.
- Drops error reporting for transient playback issues: a buffer underrun no longer changes state, so clients don't report temporary audible/visual sync hiccups.

Updates the `client/state` enum, the Playback Synchronization rules, and the handshake sequence diagram (send first `client/state` only after time synchronization is established).

Ref #88 (addresses point 3 by dropping the `error` state on underflow)

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,7 +547,7 @@ Sent once the client is ready to report its operational `state`, and whenever an
 The initial message MUST include all state fields. In subsequent messages, the client MAY send only the fields that have changed; the server MUST merge each update into existing state, retaining the last value of any field that is absent. A client MAY instead resend unchanged fields, up to its full state.
 
 - `state`: 'synchronized' | 'external_source' - operational state of the client
-  - `'synchronized'` - client is operational and ready to participate in playback; for a player this means its clock is synchronized with the server. The server does not send timestamped binary data to a client until that client reports this state.
+  - `'synchronized'` - client is operational and ready to participate in playback; for a player this means its clock is synchronized with the server. The server MUST NOT send timestamped binary data to a client until that client reports this state.
   - `'external_source'` - client is in use by an external system and is not currently participating in Sendspin playback with this server. See [External Source Handling](#external-source-handling)
 - `player?`: object - only if client has `player` role ([see player state object details](#client--server-clientstate-player-object))
 
--- a/README.md
+++ b/README.md
@@ -542,12 +542,12 @@ For synchronization, all timing is relative to the server's monotonic clock. The
 
 Client sends state updates to the server. Contains client-level state and role-specific state objects.
 
-Sent once the client is ready to report its operational `state`, and whenever any state changes thereafter. A player sends its first `client/state` only after it has established [clock synchronization](#clock-synchronization). When a role becomes active in `active_roles`, send its full state.
+Sent once the client is ready to report its operational `state`, and whenever any state changes thereafter. A player sends its first `client/state` only after it has established [clock synchronization](#clock-synchronization). The server MUST NOT send binary data to a client before that client has sent its initial `client/state`. When a role becomes active in `active_roles`, send its full state.
 
 The initial message MUST include all state fields. In subsequent messages, the client MAY send only the fields that have changed; the server MUST merge each update into existing state, retaining the last value of any field that is absent. A client MAY instead resend unchanged fields, up to its full state.
 
 - `state`: 'synchronized' | 'external_source' - operational state of the client
-  - `'synchronized'` - client is operational and ready to participate in playback; for a player this means its clock is synchronized with the server. The server MUST NOT send timestamped binary data to a client until that client reports this state.
+  - `'synchronized'` - client is operational and ready to participate in playback; for a player this means its clock is synchronized with the server.
   - `'external_source'` - client is in use by an external system and is not currently participating in Sendspin playback with this server. See [External Source Handling](#external-source-handling)
 - `player?`: object - only if client has `player` role ([see player state object details](#client--server-clientstate-player-object))
 
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ Binary audio messages contain timestamps in the server's time domain indicating 
 
 Each [`server/time`](#server--client-servertime) response provides the four timestamps needed by the filter: the client's transmitted timestamp, the server's received timestamp, the server's transmitted timestamp, and the client's receive time (captured locally when the response arrives). Clients feed these into the time filter via its `update` method and use its `compute_client_time` method to convert server timestamps to local clock values for playback scheduling.
 
-A player MUST send a `client/state` message with `state: 'synchronized'` once its time filter has converged enough to begin scheduling playback.
+A player MUST NOT report `state: 'synchronized'` until its time filter has converged enough to begin scheduling playback.
 
 ## Playback Synchronization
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,7 +172,8 @@ Each [`server/time`](#server--client-servertime) response provides the four time
 
 - Each client is responsible for maintaining synchronization with the server's timestamps
 - Clients maintain accurate sync by adding or removing samples using interpolation to compensate for clock drift
-- When a client cannot maintain sync (e.g., buffer underrun), it should send `state: 'error'` via [`client/state`](#client--server-clientstate), mute its audio output, and continue buffering until it can resume synchronized playback, at which point it should send `state: 'synchronized'`
+- On connect a client reports `state: 'not_synchronized'` via [`client/state`](#client--server-clientstate), then `state: 'synchronized'` once clock synchronization is established. The server MUST NOT send binary messages to a client unless it is in the `synchronized` state
```
maximmaxim345: Hmm, the edge case where the state changes from `synchronized` to `not_synchronized` mid stream is a bit loosely defined.
Specifically the interpretation of 'MUST NOT send binary messages to a client unless it is in the `synchronized` state' I'm worried about is: Lets just drop all those messages.

README.md
```diff
@@ -324,9 +328,9 @@ Must be sent immediately after receiving [`server/hello`](#server--client-server
 
 For the initial message, include all state fields. For subsequent updates, only include fields that have changed. The server will merge these updates into existing state.
 
-- `state`: 'synchronized' | 'error' | 'external_source' - operational state of the client
-  - `'synchronized'` - client is operational and synchronized with server timestamps
-  - `'error'` - client has a problem preventing normal operation (unable to keep up, clock sync issues, etc.)
+- `state`: 'synchronized' | 'not_synchronized' | 'external_source' - operational state of the client
+  - `'synchronized'` - client's clock is synchronized with the server; ready to receive timestamped binary data
```
maximmaxim345: Ok, thats a bit of a stretch since Music Assistant/aiosendspin doesn't send them in advance, but what about other timestamped messages (like metadata)?
I mean artwork also is about as time critical as metadata, why handle them differently?

README.md
maximmaxim345: Should we define what it means to have a 'synchronized clock`? I mean we already require the exact clock filter implementation (and with #104 also we recommend/require how to feed the time filter).

README.md
```diff
@@ -535,7 +539,7 @@ When [`stream/clear`](#server--client-streamclear) includes the player role, cli
 
 ### Server → Client: Audio Chunks (Binary)
 
-Binary messages should be rejected if there is no active stream.
+Binary messages SHOULD be rejected if there is no active stream or the client is not in the `synchronized` [state](#client--server-clientstate).
```
maximmaxim345, line 1203: Same edge case is also difficult on the client side.
Let's assume a scenario where the client just switched from `synchronized` to `not_synchronized`. But Just at this moment the server already sent the buffer for the next 30 seconds.
Even if the client switched back to `synchronized` immediately after that, it's already too late and those 30s are lost, resulting in silence.

Ok, those are a couple of tight requirements, but on the other side I think the most likely point in time where it switches to `not_synchronized` is at the very beginning of the connection, just after it became `synchronized` (when the time filter is just on the edge between synced and not synced).

maximmaxim345 (commented):
Two ideas I have for solving this edge case are:
- only allow the `not_synchronized` state at the very start
- or loosen the requirements 'The server MUST NOT send binary messages ...' and 'Binary messages SHOULD be rejected if ....' if the clock isn't synced yet the client can still buffer the data.

README.md
kahrendt: I was thinking a client can never go from `synchronized` to `not_synchronized`, but I realize I didn't spell that out. This was mainly to just guard against getting overwhelmed on initial startup.

README.md
kahrendt: Metadata messages are relatively small and are only sent once per song. Visualizer/audio send a lot of data frequently, and artwork sends a very large chunk of data all at once. That was the intention behind just blocking binary messages. Otherwise it would be a bit more convoluted to say the server can't send messages that have a timestamp to them, but I guess that's doable.

kahrendt: What about just dropping the `not_synchronized` state altogether? And we loosen the rule that the first message after the handshake is `client/state`. The client only sends that once it's synchronized. Until that stage, the server should refrain from sending binary messages (or we could even just restrict it that the server should send no messages besides time message responses)

README.md
kahrendt: Dropped `not_synchronized`

README.md
kahrendt: Dropped `not_synchronized`

README.md
kahrendt: Dropped `not_synchronized`

README.md
```diff
@@ -544,13 +542,12 @@ For synchronization, all timing is relative to the server's monotonic clock. The
...
 
 The initial message MUST include all state fields. In subsequent messages, the client MAY send only the fields that have changed; the server MUST merge each update into existing state, retaining the last value of any field that is absent. A client MAY instead resend unchanged fields, up to its full state.
 
-- `state`: 'synchronized' | 'error' | 'external_source' - operational state of the client
-  - `'synchronized'` - client is operational and synchronized with server timestamps
-  - `'error'` - client has a problem preventing normal operation (unable to keep up, clock sync issues, etc.)
+- `state`: 'synchronized' | 'external_source' - operational state of the client
+  - `'synchronized'` - client is operational and ready to participate in playback; for a player this means its clock is synchronized with the server. The server MUST NOT send timestamped binary data to a client until that client reports this state.
```
maximmaxim345: > The server MUST NOT send timestamped binary data to a client until that client reports this state.

Do we still need to bind this specifically to the `synchronized` state?
Or is this required to guard a client from being overwhelmed by data while it is streaming something else from the `external_source`?

kahrendt: https://github.com/Sendspin/spec/pull/102/commits/b00ddca70ba6187ba0a9a96ac81a15dbc67622f4

Cleaned up the gate to just be until client/state is sent. I think the external source stuff is already sufficiently covered.

README.md
kahrendt: This opens a can of worms. Do we just define it as it can translate server time to client time? Or do we specify it's accurate enough that we can meet the playback syncing standards recently added to the spec (well that specifically says we need to be accurate to within the time filter, not to absolute accuracy, so it's not quite applicable).

kahrendt: Partially addressed with https://github.com/Sendspin/spec/pull/102/commits/c8b83aaf964c2b9391016f1b4bb0150e88d926b8 Still kept a bit vague but does give something more concrete on when a client should consider itself `synchronized`

README.md
```diff
@@ -286,6 +286,8 @@ Binary audio messages contain timestamps in the server's time domain indicating
 
 Each [`server/time`](#server--client-servertime) response provides the four timestamps needed by the filter: the client's transmitted timestamp, the server's received timestamp, the server's transmitted timestamp, and the client's receive time (captured locally when the response arrives). Clients feed these into the time filter via its `update` method and use its `compute_client_time` method to convert server timestamps to local clock values for playback scheduling.
 
+A player MUST send a `client/state` message with `state: 'synchronized'` once its time filter has converged enough to begin scheduling playback.
```
maximmaxim345: Not sure if we should handle this case, but what if the just connected player has a converged time filter, but is still playing a `external_source`?
With this MUST I interpret this as: the player must lie and report `synchronized`, and after that send another `client/state` with `state: 'external_source'`.

maximmaxim345: Handling of `external_source` in general is still a bit underdefined.
Like: can a server "pull back" a client with the `external_source` state by sending a `stream/start`?
I think it's fine to merge this as is and do another PR that fixes the gaps with the `external_playback` state in one go.

Let me know if I should take a shot at it.

maximmaxim345 (approved):

README.md
kahrendt: Negated the statements here so this edge case isn't a problem: https://github.com/Sendspin/spec/pull/102/commits/e47e084d4f4249da876234947b446c2388cba63b

maximmaxim345 (approved):
Looks good now!

PR #103, opened 2026-06-15
Add seek support to the controller role

### LINKED ISSUE #99: Seek support
Request to add support for **seeking** (changing playback position within a track) to the spec.

Specifically looking for absolute position changes. As a plus, could also include relative changes (+- 5/10/30s...).

Seek probably relies on position metadata for a useable timeline, so seek is a little messier than other transport controls when determining what role to slot it into.

As a user I'm OK with cross-role requirements for this specifically. As in, adding seek to the controller role but gating behind available metadata. But can be convinced otherwise (as in, just slotting it into a metadata role)

Discussion here: https://discord.com/channels/753947050995089438/1369735163629801532/1514340390042468574

kahrendt:
Adds optional seek commands to the `controller` role. Closes #99.

## Changes

`client/command` controller object:
- `command` gains `'seek'` and `'seek_relative'`
- `position_ms` (absolute target, ms) and `offset_ms` (signed offset, ms)

`server/state` controller object:
- `supported_commands` gains `'seek'` and `'seek_relative'`, advertised independently
- `seek_max_ms` — upper bound for absolute seeks; required when `'seek'` is offered

Behaviour: `seek` is ignored if `position_ms` is outside `0..seek_max_ms`; `seek_relative` is best-effort and clamps to the seekable range.

## Notes

- A server with an unknown seekable range (e.g. live streams) can offer `'seek_relative'` without `'seek'`.
- No metadata-role dependency: `seek_max_ms` bounds absolute seek and relative seek is best-effort, so neither needs the client to know the current position.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ sequenceDiagram
     end
 
     alt Controller role
-        Client->>Server: client/command (controller: play/pause/volume/switch/etc)
+        Client->>Server: client/command (controller: play/pause/seek/volume/switch/etc)
     end
 
     alt State changes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -609,11 +613,12 @@ For clients **without** the `player` role, the cycle includes:
 The `controller` object in [`server/state`](#server--client-serverstate) has this structure:
 
 - `controller`: object
-  - `supported_commands`: string[] - subset of: 'play' | 'pause' | 'stop' | 'next' | 'previous' | 'volume' | 'mute' | 'repeat_off' | 'repeat_one' | 'repeat_all' | 'shuffle' | 'unshuffle' | 'switch'
+  - `supported_commands`: string[] - subset of: 'play' | 'pause' | 'stop' | 'next' | 'previous' | 'volume' | 'mute' | 'repeat_off' | 'repeat_one' | 'repeat_all' | 'shuffle' | 'unshuffle' | 'switch' | 'seek' | 'seek_relative'
   - `volume`: integer - volume of the whole group, range 0-100
   - `muted`: boolean - mute state of the whole group
   - `repeat`: 'off' | 'one' | 'all' - repeat mode: 'off' = no repeat, 'one' = repeat current track, 'all' = repeat all tracks (in the queue, playlist, etc.)
   - `shuffle`: boolean - shuffle mode enabled/disabled
+  - `seek_max_ms?`: integer - maximum absolute position in milliseconds a 'seek' may target (e.g., the end of the current track). The server MUST include this when 'seek' is in `supported_commands`, and MUST omit 'seek' when the seekable range is unknown (e.g., live streams); 'seek_relative' MAY still be offered
```
maximmaxim345, line 621: The only edge case of this PR that I can think of is seekable livestreams. They have an always growing `seek_max_ms`. Servers could technically start spamming clients with constant `seek_max_ms` updates. Really don't think though that we should handle this since any sane implementation would throttle this.

maximmaxim345 (approved):
LGTM.

Issue #89, opened 2026-05-27
Clock synchronization: mandate drift-SNR gate, specify burst cadence and bootstrap


balloob:
_This issue comes out of a cross-SDK conformance audit comparing every Sendspin client/server implementation (`aiosendspin`, `sendspin-cli`, `sendspin-cpp`, `sendspin-dotnet`, `sendspin-go`, `sendspin-js`, `sendspin-jvm`, `sendspin-rs`, `SendspinKit`) against the spec. **`sendspin-cpp` is treated as the reference implementation** throughout — it most closely matches the spec's language and is the only SDK that runs on the constrained-embedded target the spec was originally written for._

The spec mandates the time-filter algorithm (2-D Kalman, offset + drift)
but is silent on burst structure, cadence, outlier rejection, how to
gate playback before convergence, and whether to apply drift when the
drift estimate itself is noisy. Those four axes are where
implementations diverge. `sendspin-cpp` is the reference here because
it is the only SDK that bundles **both** ends of the cadence:
`SendspinTimeBurst` (8 messages per burst, best-RTT, 10 s) and
`SendspinTimeFilter` (drift-SNR gate, 3.0 adaptive cutoff).

### 1. Mandate the drift-SNR gate

> Clients SHOULD only apply the Kalman drift term to server→local
> conversions when `drift² ≥ k² × drift_covariance` for some `k ≥ 2`.
> Below that threshold, clients SHOULD use offset-only conversion.

This codifies what cpp/go/dotnet/js/jvm/SendspinKit already do and
forces `aiosendspin` and `sendspin-rs` to converge. Without the gate,
under-converged drift propagates wrong timestamps into the audio
scheduler.

### 2. Specify burst structure as the recommended cadence

> Clients SHOULD send `client/time` messages in bursts of 8, 10 s apart,
> and select the response with the lowest measured RTT as the burst's
> single feed into the filter. Median-of-3 selection is an acceptable
> alternative.

`sendspin-cpp` does this exactly; `sendspin-jvm` and `sendspin-js`
approximate it. The other SDKs leave cadence to the embedder, so e.g.
`sendspin-cli` runs `aiosendspin` at a fixed 10 s **single-message**
cadence — no burst denoising at all.

### 3. Standardize the adaptive cutoff

Current values vary 4×: cpp/go use 3.0, js uses 2.0,
aiosendspin/cli/jvm/rs/SendspinKit use 0.75. Either:

[...]

PR #104, opened 2026-06-16
Define minimum sync standards for the `player@v1` role

maximmaxim345:
The specification didn't exactly define a minimum bar for synced playback. That meant that terribly out-of-sync players were still valid, while their end-user experience was unusable when grouped with other Sendspin players.

The new rules:

- require the use of time-filter and the bursting strategy
- `client/time` cadence floor
- inaudible corrections in steady state
- max ±0.5% speed in steady state (sliding average over the maximum chunk size, 150 ms)
- ±1 ms accuracy in steady state
- a rare one-shot resync (startup, underrun, large disturbance) is exempt from the speed and accuracy bounds
- no startup warble
- server chunk duration bounded to 15-150 ms (covers Opus at 20 ms and FLAC at 105 ms)

To give a head start for new implementations, this PR also adds a simple suggested strategy: discrete, bit-exact sample deletion and insertion. The player drops or duplicates whole frames to correct drift, leaving the audio untouched except at the moments it corrects. N scales with sample rate. Other algorithms like ASRC are still encouraged.

## Constants

- Maximum speed deviation ±0.5%: tighter than the cli's old ±4% warble; looser than what cpp/cli hold today (~0.1%/~0.2%). This is a ~8.6 cent pitch shift, on the edge of being inaudible with music. In steady state pitch tracks clock drift, so this cap is rarely reached.
- Accuracy floor ±1 ms: achievable continuously by native clients and the Python implementation `sendspin-cli`.
- Accuracy target ±0.5 ms: in-room target, enough so individual speakers are not discernible when grouped.
- Chunk duration bounds 15-150 ms: the 150 ms cap gives headroom over `aiosendspin`'s current 105 ms max (FLAC at 44.1 kHz). The 15 ms floor keeps enough samples per chunk to correct within the ±0.5% cap.
- Soft correction baseline ≈21 µs per chunk: one frame at 48 kHz, scaled by sample rate. Implementations may use a larger step to keep up with drift, bounded by the ±0.5% cap.
- Dead band 100 µs: matches sendspin-cpp's existing value.
- One-shot resync threshold 1 ms: set equal to the accuracy floor so the snap fires before the floor is violated.

## Related issues

- #88
- #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,44 +318,25 @@ Each client is responsible for maintaining its own synchronization with the serv
 
 ### Suggested correction strategy
 
-This is one valid correction strategy for clients with the `player` role: a per-chunk push model. It is an example, not a requirement. New implementers can use it as a starting point, especially on platforms where CPU or memory usage is limited.
+This is one valid correction strategy for clients with the `player` role: discrete sample deletion and insertion. It is an example, not a requirement. New implementers can use it as a starting point, especially where CPU or memory is limited: it needs no interpolation and leaves the audio bit-exact except at the moments it corrects.
 
-Other strategies are allowed as long as they meet the rules in this section.
+Other strategies are allowed and encouraged as long as they meet the rules in this section. For example, asynchronous sample-rate conversion (ASRC) continuously resamples the stream to track the clock, trading CPU/DSP load for lower steady-state distortion than discrete frame drops.
 
-#### Per-chunk push model
+#### Sample deletion and insertion
 
-This strategy combines two mechanisms: **soft sync** handles small steady-state drift in approximately 21 µs increments per chunk, and **hard sync** handles larger one-shot errors. A "frame" is one sample across all channels (e.g. one stereo pair).
+The player renders decoded frames at their server timestamps translated to local time by the time-filter, and corrects accumulated drift by occasionally deleting or duplicating whole frames. At realistic clock drift these corrections are small and infrequent (a few per second) and individually inaudible. A "frame" is one sample across all channels (e.g. one stereo pair).
 
-**Soft sync.** Per decoded chunk:
+**Soft correction.** Per decoded chunk:
 
-1. Measure the time error between when the chunk is scheduled to play (the chunk's server timestamp translated to local time via the time-filter) and where the renderer will reach the chunk in its output buffer.
-2. If the absolute error is below the soft-sync dead band (100 µs), pass the chunk through unmodified.
-3. Otherwise, apply approximately 21 µs of correction at the chunk's end by inserting (positive error) or dropping (negative error) `N` frames, where `N = round(21 µs × sample_rate_hz / 1,000,000)`. At common sample rates: N=1 at 44.1 kHz and 48 kHz, N=2 at 96 kHz, N=4 at 192 kHz.
-4. Output the modified chunk. Residual error beyond ~21 µs carries over to subsequent chunks.
+1. Measure the time error between when the chunk is scheduled to play (its server timestamp via the time-filter) and where the renderer will reach it in the output buffer.
+2. If the absolute error is below the dead band (~100 µs), output the chunk unchanged.
+3. Otherwise correct by `N` frames: if playback is running late (the chunk reaches the output after its scheduled local time), drop `N` frames to catch up; if running early, duplicate `N` frames to wait. Residual error beyond the step carries to the next chunk.
 
-**Insert N frames at the chunk's end.** Let the chunk consist of frames `F[0], F[1], ..., F[L-1]` with `L ≥ 2`. Generate N intermediate frames between `F[L-2]` and `F[L-1]` using linear interpolation. For each `i` from 1 to N, per channel `c`:
+**Choosing N.** Use the smallest `N` that keeps up with drift, scaled to hold the step duration constant across sample rates: `N = round(21 µs × sample_rate_hz / 1,000,000)` (N=1 at 44.1 and 48 kHz, 2 at 96 kHz, 4 at 192 kHz). A chunk's correction MUST NOT exceed the ±0.5% speed cap, so `N ≤ floor(0.005 × samples_in_chunk)`. Keep `N` small; at realistic drift any `N` in this range stays masked.
 
-```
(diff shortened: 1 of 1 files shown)


maximmaxim345:
The limits/constants even work for tricky platforms like `sendspin-js` running on Chromium (which randomizes the clock for security AFAIK). It still held within about +-1 ms there.

This works because we define accuracy from the Kalman filter to the output, not end-to-end. And since we also define how the filter has to be implemented and used, this essentially factors out the things we can't control, like network delay and stability.

So even listening through a VPN across the world, it might not be perfectly in sync, but the implementation is still spec compliant, since it's doing as well as it can.

README.md
```diff
@@ -162,23 +162,64 @@ Roles with expanded allocations have slots 0-7.
...
+- **Maximum speed deviation:** The effective playback speed MUST stay within ±0.5% of normal speed, measured as a sliding average over 150 ms. This bounds continuous (steady-state) correction. A discrete one-shot resynchronization after a disturbance (startup, buffer underrun, or an error too large to correct smoothly) is not a speed deviation and is exempt; such events MUST be rare.
+
+### Sync Accuracy
+
+Sync accuracy is measured at the audio output, against what the time-filter predicts the local time should be (not against the true server clock). Use of the [time-filter](#clock-synchronization) is required to meet these minimum standards. The error is the absolute difference between when a sample actually plays in the client's local clock and the local time the time-filter predicts for that sample's server timestamp.
+
+Each client is responsible for maintaining its own synchronization with the server's timestamps.
+
+- **Accuracy floor:** In steady state, implementations MUST keep this error within ±2 ms. The only exception is the one-shot resynchronization exempted from the speed cap above, which MUST be rare.
```
kahrendt: 2 ms seems like a too high bound. Or are we limited in JS to consistently getting less than 1 ms?

README.md
```diff
@@ -162,23 +162,64 @@ Roles with expanded allocations have slots 0-7.
...
+- **Accuracy target:** Implementations SHOULD aim for ±1 ms.
```
kahrendt: Similarly, are we just limited by the JS implementation from being stricter here?

README.md
```diff
@@ -162,23 +162,64 @@ Roles with expanded allocations have slots 0-7.
...
+
+### Startup Behavior
+
+- **No startup warble:** During startup, the client MUST NOT produce audible pitch modulation, warble, or other transient artifacts in the audio output.
+
+### Server Audio Send Constraints
+
+- **Chunk duration bounds:** A server MUST NOT send an audio chunk longer than 150 ms, and SHOULD NOT send one shorter than 15 ms (the final chunk of a stream or the chunk before a format change MAY be shorter).
+- The server sends audio to late-joining clients with future timestamps only, allowing them to buffer and start playback in sync with existing clients.
+- After sending [`stream/start`](#server--client-streamstart) or [`stream/clear`](#server--client-streamclear) messages, servers must schedule the first audio timestamp far enough in the future to satisfy each player's [`required_lead_time_ms`](#client--server-clientstate-player-object) (startup warmup) and [`min_buffer_ms`](#client--server-clientstate-player-object) (ongoing jitter buffer). For live streams the buffer cannot grow after playback begins, so the larger of the two must already be reached before the first chunk plays.
```
kahrendt, line 316: For a live stream, why would we take the larger of the two (not exactly this PR but a question that's been bugging me as I implement the player timing changes!)?

Presumably, on a live stream, you don't care about missing the first chunk of audio because... it's live, you've already missed the ongoing stuff. The network jitter is really the only one to care about in that scenario; i.e., how much buffer do I need to avoid audio drops?

README.md
maximmaxim345: `sendspin-js` shouldn't be limiting the spec so we can lower it.
What about ±1ms? I wouldn't go lower without more testing though. From some initial tests ±0.5ms looks difficult to achieve with a USB DAC playing with `sendspin-cli`.

README.md
maximmaxim345: How about ±0.5ms? It's a recommendation so we could technically go as low as we want. It what implementations target after all, if they run on slow hardware it just might be slightly above the threshold (but still compliant).

README.md
maximmaxim345: `required_lead_time` isnt strictly about the missing content, its about how much time should happen between the `stream/start` and the timestamp in the first chunk. So to have `required_lead` and `min_buffer` work consistently for the player (no matter if its live or not), taking the larger of the two should make it work without having a buffer.

Buffered can of course ignore the `min_buffer` since they can just fetch more audio data. But with live streams only using `required_lead` or only using `min_buffer` would cause it to break one way or the other. If you only use `required_lead` and its smaller than the `min_buffer`, the buffer starts too shallow and since live cant grow it after playback starts you get dropouts the first time the network jitters. And if you only use `min_buffer` and its smaller than `required_lead`, then the first chunk lands less than `required_lead` after `stream/start`, which contradicts the spec since thats exactly the gap its supposed to guarantee.

kahrendt: Then why do we have the two different parameters if we just take the max? If a client has a slow startup but rock solid network, I thin it would more useful to use just the network jitter measurement, even if that means the client needs to throw away a few chunks at the start (which is a violation of the spec as written, to be fair).

Let's see if I'm understanding by phrasing it a bit differently. If it isn't a live stream, then only `required_lead` is honored. If it is a live stream, then we take the max of `required_lead` and `min_buffer`. Which just effectively means the `min_buffer` parameter only matters if `min_buffer > required_lead` in the live case. Do I have that breakdown right for how the server behaves? I feel like we are missing an opportunity for the other case.

We can move this to Discord or a separate discussion so I don't keep polluting this PR!

README.md
kahrendt: Let's drop this to plus or minus 1 ms.

README.md
kahrendt: 0.5ms is fine!

kahrendt (approved):

README.md
kahrendt: Addressed in https://github.com/Sendspin/spec/pull/107

PR #105, opened 2026-06-29
Add `source` role for audio input clients

maximmaxim345:
Adds the `source@v1` role that allows a client to capture local audio (line-in, turntable, Bluetooth receiver, microphone) and stream it to the server, while the server stays the single place that resamples, transcodes, mixes, buffers, and distributes to players (including back to the source itself when it is also a player, so it stays in sync).

Heavily based on #52 by @rudyberends but simplified to the core. This PR specifically removes:
- the activation model (input type, activation kind)
- `level` reporting (kept `signal` line-sensing)
- the `client/command` source notifications
- `select_source` controller integration and source listing in `server/state`
- server-side virtual source clients

Also rebases it onto the encryption work (#84), and now requires source clients to be paired since an unpaired source could otherwise inject malformed audio into the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1194,7 +1194,6 @@ The `source@v1_support` object in [`client/hello`](#client--server-clienthello) 
 The `source` object in [`client/state`](#client--server-clientstate) has this structure:
 
 - `source`: object
-  - `state`: 'idle' | 'streaming' - whether the source is currently capturing and sending audio
   - `signal?`: 'present' | 'absent' - optional line sensing/signal presence, only if 'line_sense' is supported
 
 ### Server → Client: `server/command` source object
@@ -1207,17 +1206,17 @@ The `source` object in [`server/command`](#server--client-servercommand) has thi
 #### Source command semantics
 
 - `command` controls Sendspin ingest lifecycle for this source:
-  - `start`: server requests ingest to become active. The client SHOULD transition to `state: "streaming"`, send `input_stream/start`, and then send source audio chunks.
-  - `stop`: server requests ingest to become inactive. The client SHOULD send `input_stream/end`, stop sending source audio chunks, and transition to `state: "idle"`.
+  - `start`: server requests ingest to become active. The client SHOULD send `input_stream/start` and then send source audio chunks.
+  - `stop`: server requests ingest to become inactive. The client SHOULD send `input_stream/end` and stop sending source audio chunks.
 
 #### Default ingest behavior
 
 - Effective default after handshake is `stop` (ingest inactive).
 - Server ingest interest is represented by `command: "start"` / `command: "stop"`.
 
-A source MAY also start ingest on its own without waiting for `command: "start"`: it transitions to `state: "streaming"`, sends `input_stream/start`, and begins sending chunks. This lets a source that detects local signal (e.g., a turntable starting) begin streaming immediately, with no `signal`-to-`start` round trip. The server infers a source-initiated start from the `input_stream/start` it did not request, decides whether to route or drop the stream per its own policy, and a source MUST still honor a subsequent `command: "stop"`.
+A source MAY also start ingest on its own without waiting for `command: "start"`: it sends `input_stream/start` and begins sending chunks. This lets a source that detects local signal (e.g., a turntable starting) begin streaming immediately, with no `signal`-to-`start` round trip. The server infers a source-initiated start from the `input_stream/start` it did not request, decides whether to route or drop the stream per its own policy, and a source MUST still honor a subsequent `command: "stop"`.
 
-A source MAY likewise stop on its own (e.g., the input goes silent): it sends `input_stream/end`, stops sending chunks, and reports `state: "idle"`.
+A source MAY likewise stop on its own (e.g., the input goes silent): it sends `input_stream/end` and stops sending chunks.
 
 When the server removes `source` from [`active_roles`](#server--client-serveractivate), the client sends `input_stream/end` and stops sending chunks.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1158,6 +1163,90 @@ Binary messages should be rejected if there is no active stream.
...
+
+- Byte 0: message type `12` (uint8)
+- Bytes 1-8: timestamp (big-endian int64) - server clock time in microseconds when the first sample was captured
+- Rest of bytes: encoded audio frame
+
+The timestamp indicates when the first audio sample in this chunk was captured (in server time domain). The server MAY resample/transcode and then distribute the audio to players with its normal buffering and synchronization strategy.
+
+A device that implements both `source` and `player` MUST NOT play its captured input locally. Like every player, it outputs only the stream the server distributes, so its output stays in sync with the rest of the group.
+
+**Note:** Source timestamps are derived from the client's clock offset, which the time filter keeps re-estimating, so they may show discontinuities or drift (e.g., ADC clock variance). Server implementations SHOULD NOT assume perfectly continuous timestamps; the audio sample stream itself SHOULD remain continuous.
```
maximmaxim345, line 1304: We should describe how the server should behave too. Like prescribing the use of ASRC and maybe how to handle network jitter as well. The protocol and client implementation stay the same though so we could solve this in a later follow up PR.

README.md
```diff
@@ -1158,6 +1163,90 @@ Binary messages should be rejected if there is no active stream.
...
 
+## Source messages
+This section describes messages specific to clients with the `source` role, which capture audio from a local input (e.g., AUX/line-in, turntable preamp, Bluetooth receiver, or microphone) and stream it to the server. Unlike other roles, a source sends audio to the server; the server remains the single place that resamples, transcodes, mixes, buffers, and distributes audio to output players. Sources stay simple: they capture and encode audio, optionally report basic signal presence (line sensing), and stream timestamped audio frames.
+
+A device MAY implement both the `source` and `player` roles (e.g., a speaker with a local AUX input forwarded into Sendspin).
+
+**Note:** The `source` role (capturing input *into* Sendspin) is distinct from the client-level [`state: 'external_source'`](#external-source-handling), which marks a client whose *output* has been taken over by a non-Sendspin system.
+
```
maximmaxim345, line 1228: We could technically rename either `external_source` or `source` if this is a problem.
IMO it's fine calling both source.

README.md
```diff
@@ -1158,6 +1163,90 @@ Binary messages should be rejected if there is no active stream.
...
+- `command` controls whether this source streams to the server:
+  - `start`: server requests the source to begin streaming. The client SHOULD send `input_stream/start` and then send source audio chunks.
+  - `stop`: server requests the source to stop streaming. The client SHOULD send `input_stream/end` and stop sending source audio chunks.
+
+#### Default streaming behavior
+
+The default after the handshake is `stop`: a source MUST NOT stream until the server sends `command: "start"`. The server is the only party that initiates streaming.
+
+A source that supports line sensing reports `signal` in [`client/state`](#client--server-clientstate). The server MAY use it as a hint for when to send `command: "start"` or `command: "stop"`, but the decision is server policy.
```
maximmaxim345, line 1272: Not sure, I think it's good to require the server to switch to the input, but I can see a couple cases where this might become annoying. Especially if it's a combined source+player that is already grouped with other players. Or if the connected input is very noisy.

README.md
maximmaxim345, line 1304: With this PR merged:
- https://github.com/Sendspin/spec/pull/102

We should also gate so clients can only send audio chunks after it's clock is `synchronized`.

README.md
```diff
@@ -1164,6 +1169,90 @@ Binary messages should be rejected if there is no active stream.
...
+
+When the server removes `source` from [`active_roles`](#server--client-serveractivate), the client sends `input_stream/end` and stops sending chunks.
+
+### Client → Server: `input_stream/start`
```
kahrendt: Is it necessary to create a new message type. Could we just use `stream/start` instead of `input_stream/start`? Is there some ambiguity that would make things confusing?

maximmaxim345: Technically not, but it's easier to search the spec and also avoids some confusion when implementing the spec.
Right now each message type is reserved for exactly one direction, So having two `stream/start` messages with a different schema would break this convention.
We have `client/time` and `server/time` messages for example.

kahrendt (approved):

kahrendt:
LGTM! Your comments can all be addressed in follow-up clarifications.

README.md
maximmaxim345: What about renaming this from `input_stream` to `client_stream`?

kahrendt: Sounds better, change it :)

kahrendt (approved):
Thanks for the message rename!

README.md
maximmaxim345: Addressed in [af29bba](https://github.com/Sendspin/spec/pull/105/commits/af29bba4ee99a3257460296d290036208122d458).

kahrendt (approved):
Changes in https://github.com/Sendspin/spec/pull/105/commits/af29bba4ee99a3257460296d290036208122d458 look good. I'll let you do the honors of merging in case you have anything else you want to touch

Hedda:
> Heavily based on #52 by @rudyberends but simplified to the core. This PR specifically removes:
> 
> * the activation model (input type, activation kind)
> * `level` reporting (kept `signal` line-sensing)
> * the `client/command` source notifications
> * `select_source` controller integration and source listing in `server/state`
> * server-side virtual source clients
> 
> Also rebases it onto the encryption work (#84), and now requires source clients to be paired since an unpaired source could otherwise inject malformed audio into the server.

Should @rudyberends maybe create new PRs for any of those functions that was removed in this PR?

PR #113, opened 2026-07-01
Resolve client-facing spec gaps

maximmaxim345:
This PR and one more (the `state` to `available` rename in https://github.com/Sendspin/spec/pull/115) are the last client-breaking changes to the spec. There *might* be more if we hit issues with the newly added encryption support, but apart from that these two *should* cover everything I found.

There are also some breaking changes I'm still working out, but those are strictly server-implementation breaking. Clients will not have to adjust anything.

Specifically, this PR addresses the following gaps:
- `client/goodbye`: a client switching servers MUST send reason `another_server`, and clients SHOULD default to `shutdown` when the device is powering off or otherwise not coming back.
- Discovery: the `path` TXT key is now REQUIRED (value still recommends `/sendspin`).
- Pairing: the initial static PIN MUST be device-specific (e.g. randomly generated and printed), never a fixed shared default.
- Custom roles: `_`-prefixed role names SHOULD carry a vendor-specific prefix to avoid collisions.
- Player state: persisting `volume`/`muted` across reboots is RECOMMENDED, and a server MUST NOT assume they are unchanged after a reconnect.
- Streaming: `codec_header` is specified as standard Base64 (the rest of the spec uses base64url), clarifying the existing encoding rather than changing the wire format.
- `external_source`: a client whose output is taken by a non-Sendspin activity reports `external_source` only when it will not yield (the server then cannot stream to it, since `stream/start` is gated on `synchronized`); a client that can be interrupted instead stays `synchronized` so the server can take over at any time.
- Source role: dropped `source@v1_support.supported_formats`; a source declares its format per stream in `client_stream/start` and the server accepts any format.
- Forward compatibility: clients MUST ignore unrecognized `payload` fields, and MUST NOT send fields the spec doesn't define for a message (beyond the `_`-prefixed application-specific role objects a message already permits). Undefined field names stay reserved to the spec, so a client can't squat on a name a later version might claim.
- `stream/request-format`: a player MUST only request a format listed in its `supported_formats`, so a format request can never name something the client cannot decode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,7 +699,7 @@ Sent by the client before gracefully closing the connection. This allows the cli
 Upon receiving this message, the server should initiate the disconnect.
 
 - `reason`: 'another_server' | 'shutdown' | 'restart' | 'user_request' | 'unauthorized' | 'pairing_required' | 'concurrent_attempt' | 'unpaired'
-  - `another_server` - client is switching to a different Sendspin server. A client that leaves one server for another MUST send this reason to the server it is leaving. Server should not auto-reconnect but should show the client as available for future playback
+  - `another_server` - client is switching to a different Sendspin server. A client that leaves one server for another MUST send this reason to the server it is leaving. Server SHOULD NOT auto-reconnect but SHOULD show the client as available for future playback
   - `shutdown` - client is shutting down. Server should not auto-reconnect
   - `restart` - client is restarting and will reconnect. Server should auto-reconnect
   - `user_request` - user explicitly requested to disconnect from this server. Server should not auto-reconnect
--- a/README.md
+++ b/README.md
@@ -569,7 +569,7 @@ A client's output can be taken over by a non-Sendspin activity (playing other me
 
 If the external activity can be interrupted by Sendspin playback at any time, the client SHOULD remain `synchronized` so the server can take it over.
 
-To stop rendering its group's audio while it does its own thing, a client MAY leave its current group by sending a `client/state` with `state: 'external_source'` immediately followed by a `client/state` with `state: 'synchronized'` (the server behavior below then moves it to a solo group and does not rejoin it). This is only needed while the group's `playback_state` is `'playing'`.
+To stop rendering its group's audio while performing non-Sendspin activity, a client MAY leave its current group by sending a `client/state` with `state: 'external_source'` immediately followed by a `client/state` with `state: 'synchronized'` (the server behavior below then moves it to a solo group and does not rejoin it). This is only needed while the group's `playback_state` is `'playing'`.
 
 #### Non-interruptible activity (client becomes unavailable)
 
--- a/README.md
+++ b/README.md
@@ -563,9 +563,19 @@ The initial message MUST include all state fields. In subsequent messages, the c
 
 ### External Source Handling
 
-When a client sets `state: 'external_source'`, it indicates the client's output is in use by an external system (e.g., a different audio source, HDMI input, or local media playback) and is not currently participating in Sendspin playback with this server.
+A client's output can be taken over by a non-Sendspin activity (playing other media, another protocol, an HDMI input, and so on). How it reports this depends on whether it will still yield its output back to Sendspin on request.
 
-A client SHOULD report `external_source` only while it will not yield its output to Sendspin. If the current external activity can be interrupted by Sendspin playback, the client SHOULD stay `synchronized` (letting the server take over at any time) rather than reporting `external_source`, and SHOULD return to `synchronized` as soon as it is again willing to be taken over.
+#### Interruptible activity (client stays available)
+
+If the external activity can be interrupted by Sendspin playback at any time, the client SHOULD remain `synchronized` so the server can take it over.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -565,6 +565,8 @@ The initial message MUST include all state fields. In subsequent messages, the c
 
 When a client sets `state: 'external_source'`, it indicates the client's output is in use by an external system (e.g., a different audio source, HDMI input, or local media playback) and is not currently participating in Sendspin playback with this server.
 
+A client SHOULD report `external_source` only while it will not yield its output to Sendspin. If the current external activity can be interrupted by Sendspin playback, the client SHOULD stay `synchronized` (letting the server take over at any time) rather than reporting `external_source`, and SHOULD return to `synchronized` as soon as it is again willing to be taken over.
```
maximmaxim345: state should be `available` and `unavailable` instead.
Or even better: a binary `available`.

TRIAGE.md
```diff
@@ -0,0 +1,36 @@
...
+
+Admission accepts an incoming connection of "higher or equal" priority (README.md:78), so an
+equal-priority server displaces the current holder. Two servers that both want, e.g., `management`
+can each displace the other on reconnect indefinitely. Backoff is mandated only for repeated
+Noise-handshake failures (README.md:699), not for this admission ping-pong.
+
+## Possibly breaking (new states/messages; largely fills a currently undefined gap)
+
+### 49. Source format is declared unilaterally, so `source@v1_support.supported_formats` has no selection path
+Confirmed: a server should accept any source format, so `source@v1_support.supported_formats` has no
```
maximmaxim345: Lets drop `source@v1_support.supported_formats`. Server should accept anything.

TRIAGE.md
```diff
@@ -0,0 +1,36 @@
+# Triage
+
+Breaking questions still open for discussion, most disruptive first.
+
+## Likely hard-breaking (client wire/behavior change if adopted)
+
+### 24. Equal-priority displacement enables server oscillation with no mandated backoff
```
maximmaxim345: Solvable by strengthening the reconnect rule. Server-side only without client impact, so moved out of this PR.
Will open a separate PR later.

README.md
maximmaxim345: I'll open a PR after this one to do the rename.

To solve indicate the: playing but you can take back control case we now:
- suggest flipping to external_source and back to synchronized
- force the server to no rejoin the player once it goes back to the `synchronized` state.

README.md
```diff
@@ -563,7 +563,19 @@ The initial message MUST include all state fields. In subsequent messages, the c
...
 ### External Source Handling
 
-When a client sets `state: 'external_source'`, it indicates the client's output is in use by an external system (e.g., a different audio source, HDMI input, or local media playback) and is not currently participating in Sendspin playback with this server.
+A client's output can be taken over by a non-Sendspin activity (playing other media, another protocol, an HDMI input, and so on). How it reports this depends on whether it will still yield its output back to Sendspin on request.
+
+#### Interruptible activity (client stays available)
+
+If the external activity can be interrupted by Sendspin playback at any time, the client SHOULD remain `synchronized` so the server can take it over.
+
+To stop rendering its group's audio while it does its own thing, a client MAY leave its current group by sending a `client/state` with `state: 'external_source'` immediately followed by a `client/state` with `state: 'synchronized'` (the server behavior below then moves it to a solo group and does not rejoin it). This is only needed while the group's `playback_state` is `'playing'`.
```
kahrendt: "To stop rendering its group's audio while performing non-Sendspin activity,"

Or something like that. I don't like the "does its own thing"

README.md
```diff
@@ -683,7 +699,7 @@ Sent by the client before gracefully closing the connection. This allows the cli
 Upon receiving this message, the server should initiate the disconnect.
 
 - `reason`: 'another_server' | 'shutdown' | 'restart' | 'user_request' | 'unauthorized' | 'pairing_required' | 'concurrent_attempt' | 'unpaired'
-  - `another_server` - client is switching to a different Sendspin server. Server should not auto-reconnect but should show the client as available for future playback
+  - `another_server` - client is switching to a different Sendspin server. A client that leaves one server for another MUST send this reason to the server it is leaving. Server should not auto-reconnect but should show the client as available for future playback
```
kahrendt: Should we be using the RFC language for "should not auto-reconnect" and "should show"?

maximmaxim345: Changed it just for this line now, but we should also apply the RFC language to the rest of the spec in another PR. (thinking after #114 so it's easier to review).

README.md
maximmaxim345: That sounds a lot better! Changed it.

README.md
kahrendt: Thanks for the fix! Yeah another pass is definitely needed, I've just been trying to use RFC language on anything I touch (#114 is definitely an exception and issues shouldn't be fixed there since it is just splitting stuff up).

kahrendt (approved):
Thanks, good clarifications!

PR #114, opened 2026-07-02
Generate `README.md` from split spec sources

maximmaxim345:
The spec used to be one big `README.md`. It's now split into per-section and per-role source files, and `README.md` is generated from them by `tools/build-readme.py`.

The generated `README.md` in this PR matches the old one except for two things the build forces: the generated-file banner at the top, and the two duplicate "Multiple servers" headings, renamed to "(server-initiated)" and "(client-initiated)" (with their two inbound links) so their anchors stop colliding. No sections move, so the diff against the old `README.md` is tiny. Reordering the page is left to a follow-up PR.

`template.md` is where the assembly lives: the document head, then `<!-- include: <path> -->` lines in page order. Links between files are written file-relative and turned into page anchors at build time, and the build fails on duplicate or dangling anchors. HTML comments are stripped from the output unless a `<!-- keep: ... -->` line precedes them, which is what keeps the banner. The pre-commit hook regenerates `README.md` from the staged files and blocks hand-edits to it. CI runs the same `--check`. Also adds `CONTRIBUTING.md`.


.github/workflows/readme-sync.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
kahrendt: Can you pin these actions to the latest SHAs?

tools/build-readme.py
```diff
@@ -0,0 +1,224 @@
...
+    """A source problem that stops the page from being generated."""
+
+
+def slug(heading):
+    h = re.sub(r"[^a-z0-9 -]", "", heading.strip().lower())
+    return h.strip().replace(" ", "-")
+
+
+def headings(text):
+    return re.findall(r"^#{1,6}\s+(.*)$", text, re.M)
```
kahrendt, line 36: This drops underscores, but GitHub preserves them. Three headings are affected: `` `client_stream/start` ``, `` `client_stream/end` `` (roles/source/v1.md) and `` `f_peak` — message type `18` `` (roles/visualizer/v1.md).

The last one gets built as `fpeak--message-type-18`; GitHub's real anchor is `f_peak--message-type-18`.

kahrendt: These aren't referenced anywhere as far as I can see, but it's good future proofing!

kahrendt (changes requested):

kahrendt (approved):
Thanks!

PR #115, opened 2026-07-02
Rename client `state` to a boolean `available`

maximmaxim345:
The remaining client-breaking change after #113: it replaces the `client/state` `state` enum (`synchronized` | `external_source`) with a boolean `available`.

- `available: true` - operational and ready to participate; for a player or source this means its clock has converged. Formerly `synchronized`.
- `available: false` - the client's output is in use by an external system. Formerly `external_source`.

Pure terminology rename with no behavioral change. The group-leave behavior it builds on (the `available: false` then `available: true` flip, and the no-auto-rejoin rule) already landed in #113.


kahrendt (approved):
Looks good!

PR #117, opened 2026-07-08
Wrap PSK under CPace output

arturpragacz:
Wrap PSK under CPace output, which is the more standard construction and gives us the benefits of full applicability of all CPace security proofs.

Also improve pin protocol error handling, by clarifying which errors result in `pin_mismatch`.

Set `ADa` and `ADb` to distinct values to fix a security issue with reflected MAC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -718,7 +718,7 @@ Upon receiving this message, the server should initiate the disconnect.
 
 ## Pairing
 
-Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](#re-handshake) to the newly delivered `long_term_psk`, bringing the channel under the new trust ceiling without closing the WebSocket.
+Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](#re-handshake) to the newly delivered PSK, bringing the channel under the new trust ceiling without closing the WebSocket.
 
 This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement either or both PIN methods.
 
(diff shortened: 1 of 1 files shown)


README.md
maximmaxim345: This section (line 721) still mentions the `long_term_psk`.

README.md
```diff
@@ -896,6 +897,22 @@ The four pairing message fields carry the corresponding CPace values, base64url-
...
+K_wrap = SHA-256("sendspin-pair-psk-wrap-v1" || sid || ISK)
+```
+
+The hash input is the UTF-8 bytes of the literal label (no separator, no NUL terminator) followed by `sid` (the CPace session id defined in [PAKE](#pake), raw) and `ISK` (the 64-byte CPace intermediate session key, raw). The client encrypts the 32-byte PSK with the AEAD of the connection's negotiated [cipher suite](#cipher-suites), key `K_wrap`, a 12-byte all-zero nonce, and empty associated data. `wrapped_psk` carries the 48-byte ciphertext-plus-tag, base64url-encoded without padding (64 chars).
+
+To unwrap, the server decrypts `wrapped_psk` with the same AEAD, key `K_wrap`, and nonce, recovering the 32-byte PSK.
+
+### Protocol Errors
+
+A condition no conformant peer produces - a malformed or missing field, a CPace share with the wrong length or encoding a low-order point, a revealed nonce that does not match its commitment, a `wrapped_psk` that fails to decrypt - is a **protocol error**: the detecting side closes the WebSocket without sending any application-level error message, and persists nothing.
```
maximmaxim345:
```suggestion
A condition during pairing that no conformant peer produces - a malformed or missing field, a CPace share with the wrong length or encoding a low-order point, a revealed nonce that does not match its commitment, a `wrapped_psk` that fails to decrypt - is a **protocol error**: the detecting side closes the WebSocket without sending any application-level error message, and persists nothing.
```

README.md
```diff
@@ -885,7 +885,8 @@ Sendspin instantiates CPace's inputs as follows:
 - `PRS` - the PIN as a UTF-8 byte string (the literal decimal digits - e.g., `0x31 0x32 0x33 0x34 0x35 0x36 0x37 0x38` for the PIN `"12345678"`).
 - `sid` - the UTF-8 bytes `"sendspin-pair-pake-v1"` concatenated with the Noise handshake hash `h` available immediately after Noise transport mode begins.
 - `CI` - empty.
-- `ADa`, `ADb` - empty.
+- `ADa` - the UTF-8 bytes `"server"`.
+- `ADb` - the UTF-8 bytes `"client"`.
```
maximmaxim345, line 889: Isn't this slightly redundant since we already prescribe message ordering between server and client?

Also, can you edit the PR description to mention that `ADa`/`ADb` changed?

maximmaxim345 (approved):
Left a small question/nit about the `ADa`/`ADb` change, but other than that this looks correct.

README.md
arturpragacz: No, it is not. We are not sending these values over the wire, which indeed would be redundant, but they partake in the crypto calculations.

In fact this is arguably the most important change in this PR from the security perspective as it fixes a real serious vulnerability (although with the other changes here it is not that exploitable, still good to have it).

PR #118, opened 2026-07-08
Add domain separator to pair commit

arturpragacz:
Add domain separator to pair commit.

We use labels for all the other constructions in the spec, it's not necessary here, but it's a good security practice to do it everywhere.


maximmaxim345 (approved):
Looks good.

PR #120, opened 2026-07-13
Keep the connection open across pairing attempts

arturpragacz:
Keep the connection open across pairing attempts.

The intention of allowing only one attempt per connection was to simplify things, but it turns out it actually makes the implementation more annoying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Clients must persistently store the `server_id` of the server that most recently
 
 When a new server connects, the client lets the handshake complete before applying admission; the new connection is provisional until its first [`server/activate`](#server--client-serveractivate) declares its priority. The incoming connection's priority is compared to the current connection's: higher or equal is accepted, lower is rejected. Two exceptions:
 
-- An in-flight pairing is not displaced by an incoming `'playback'` or `'pairing'` connection.
+- A [pairing attempt](#entering-and-leaving-pairing) is not displaced by an incoming `'playback'` or `'pairing'` connection.
 - When both the current holder and the incoming connection have empty `activities`, the incoming is admitted only if its `server_id` matches the last-playback server (and the existing one's does not); otherwise the existing is kept.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -883,7 +883,7 @@ The PIN pairing flows use **CPACE-X25519-SHA512** as the PAKE construction, defi
 Sendspin instantiates CPace's inputs as follows:
 
 - `PRS` - the PIN as a UTF-8 byte string (the literal decimal digits - e.g., `0x31 0x32 0x33 0x34 0x35 0x36 0x37 0x38` for the PIN `"12345678"`).
-- `sid` - the UTF-8 bytes `"sendspin-pair-pake-v1"` concatenated with the Noise handshake hash `h` available immediately after Noise transport mode begins.
+- `sid` - the UTF-8 bytes `"sendspin-pair-pake-v1"` || `h` || `counter`. `h` is the Noise handshake hash (32 bytes, raw) available immediately after Noise transport mode begins; `counter` is the number of pairing [`server/activate`](#server--client-serveractivate) messages sent since the last Noise handshake (`0` for the first attempt), encoded as a big-endian uint32 (4 bytes).
```
maximmaxim345: Wouldn't the number of `server/activate` messages be 1 for the first attempt?

arturpragacz: Does the array start from index 1 or 0? ;)

But joking aside, it obviously doesn't matter. Do you wish to change it here to start from 1?

maximmaxim345: Haha, starting from 0 is good 😀
My point is that "`counter` is the number of pairing `server/activate` messages" and "`0` for the first attempt" sounds like a slight contradiction. Second attempt could technically be interpreted as a 2 (the initial and second `server/activate` messages).
Same wording is in `client/pair-init`.

arturpragacz: Ok, let's count from 1, that seems like the simplest solution.

maximmaxim345 (approved):

PR #122, opened 2026-07-16
Resolve more pre-1.0 spec gaps

maximmaxim345:
Resolves a batch of pre-1.0 ambiguities across the spec.

- **Roles and preamble:** codec framing, source format and trust rules, group volume and mute for mixed and empty groups, artwork channel and lateness semantics, the metadata clock domain, normative language boilerplate, and role versioning notes.
- **Connection and management:** plain `ws://` transport, discovery name precedence, provisional connection caps, prologue hashing, `psk_id` uniqueness across PSK categories (collisions rejected as `already_exists`), and a device-specific pre-provisioned fallback record.
- **Messaging:** WebSocket control frames vs the binary-only rule, per-direction fragmentation and malformed-sequence handling, handshake payload encodings and `version`/`suite` typing, re-handshake framing and key-swap boundary, required role support objects and `supported_pair_methods`, per-role trust folded into the central `server/activate` rules, state merge depth and initial full-state baselines, and `client/goodbye` reason cleanup.
- **Pairing:** PIN pairing over an already-paired channel, stream-only quiesce when entering pairing, server-side cancel and timeout, Pairing PSK lifecycle and standing client obligations, a pinned CPace draft revision, exact lockout counter events, and sequence violations handled as protocol errors.
- **Source role:** prompt and idempotent `start/stop`, clarify per-connection streaming state, chunks in flight after a stop, capture timestamps under encoder delay, chunk duration bounds and stall backlog, and rate estimation from the delivered sample stream.

Breaking: the group volume/mute formula, which fixes edge cases for groups with no volume- or mute-capable members. Breaking for servers only.

Three `client/hello` and `client/state` rules are formally new MUSTs but match what existing implementations already do, so they carry no practical adoption cost: a role's support object (e.g. `player@v1_support`) is now required whenever that role version is listed in `supported_roles`, `supported_pair_methods` is now required (it can never be legitimately absent, every client implements at least Pairing PSK), and clients with only display stream roles (`artwork`, `visualizer`) now also send the initial `client/state`.


management.md
```diff
@@ -34,6 +34,8 @@ Add a pairing record directly.
 - `psk`: string - 43-character base64url-encoded 32-byte [Sendspin PSK](README.md#definitions) (no padding)
 - `server_id?`: string - present for stored-pubkey records, absent for shared-PSK records
 
+A `psk` whose `psk_id` collides with a candidate PSK in a different category (the Sentinel PSK or the client's Pairing PSK; see [Pre-Shared Key](connection.md#pre-shared-key)) is rejected as `invalid`. The same PSK already present as a record is `already_exists`.
```
arturpragacz: I think returning `already_exists` is more useful here?

management.md
```diff
@@ -96,9 +98,12 @@ Modify pairing config.
 - `unpaired_access?`: object - see [Unpaired Access](pairing.md#unpaired-access)
   - `enabled?`: boolean
 
-The request applies as a patch: only fields present in the payload are written, and any absent field (including an absent method object) leaves the corresponding stored value unchanged. Setting fields on a method the client does not implement returns `invalid`.
+The request applies as a patch: only fields present in the payload are written, and any absent field (including an absent method object) leaves the corresponding stored value unchanged. Two cases are rejected as `invalid`:
+
+- fields set on a method the client does not implement
+- a `pairing_psk.psk` whose `psk_id` collides with a candidate PSK in a different category (the Sentinel PSK or a stored record; see [Pre-Shared Key](connection.md#pre-shared-key))
```
arturpragacz: Same as above.

kahrendt (approved):
Looks good to me!

messaging.md
```diff
@@ -144,12 +151,14 @@ Carries one Noise handshake message. Sent twice during the handshake: once by th
...
+
+A malformed inner handshake payload (not valid UTF-8 JSON of the shape above) is a handshake failure and closes the WebSocket (see [Failure Handling](connection.md#failure-handling)).
 
-After both handshake messages have been exchanged, both sides switch to Noise transport mode. All subsequent WebSocket frames are binary, and all payloads are Noise transport ciphertexts.
+After both handshake messages have been exchanged, both sides switch to Noise transport mode (all subsequent messages travel as the binary Noise-ciphertext frames described above).
 
-The same `noise/handshake` message is used for the in-band [re-handshake](connection.md#re-handshake): the two messages then travel as binary frames encrypted under the current transport keys rather than as cleartext text frames.
+The same `noise/handshake` message is used for the in-band [re-handshake](connection.md#re-handshake): the two messages then travel as ordinary encrypted JSON messages (binary frames, message type `0`), not bare Noise bytes. Noise message 2 is still encrypted under the pre-re-handshake transport keys; the first frame each side sends after performing `split()` uses the new keys.
```
arturpragacz: I would not mention `split()` here, maybe just say that all subsequent messages use new keys.

messaging.md
```diff
@@ -178,25 +187,27 @@ Players that can output audio should have the role `player`.
...
+- `player@v1_support?`: object - required if `player@v1` is listed, absent otherwise ([see player@v1 support object details](roles/player/v1.md#client--server-clienthello-playerv1-support-object))
+- `source@v1_support?`: object - required if `source@v1` is listed, absent otherwise ([see source@v1 support object details](roles/source/v1.md#client--server-clienthello-sourcev1-support-object))
+- `artwork@v1_support?`: object - required if `artwork@v1` is listed, absent otherwise ([see artwork@v1 support object details](roles/artwork/v1.md#client--server-clienthello-artworkv1-support-object))
+- `visualizer@v1_support?`: object - required if `visualizer@v1` is listed, absent otherwise ([see visualizer@v1 support object details](roles/visualizer/v1.md#client--server-clienthello-visualizerv1-support-object))
+- `supported_pair_methods`: object[] - pairing methods this client offers, each described by a [pair-method descriptor](pairing.md#client--server-clienthello-pair-method-descriptor). Never empty: every client implements at least the Pairing PSK method (see [Pairing](pairing.md#pairing)).
```
arturpragacz: Technically can be empty, if all are disabled.

pairing.md
```diff
@@ -28,16 +33,20 @@ The same `server/activate` can also end a pairing attempt without finalizing: se
...
 
 ### Pairing PSK Flow
 
 The Noise handshake completes using the Pairing PSK, authenticating both sides. The client proceeds straight to [`client/pair-finalize`](#client--server-clientpair-finalize).
 
+**Lifecycle.** The client generates its Pairing PSK from a CSPRNG and persists it across reboots. It is per-client and long-lived: a successful pairing does not consume or rotate it (pairing produces a separate long-term [Sendspin PSK](README.md#definitions)), so it can pair the client with any number of servers. The client MAY rotate it at any time, and a paired server can rotate it via [`management/set-pairing-config`](management.md#server--client-managementset-pairing-config) (`pairing_psk.psk`); rotation invalidates previously distributed copies but leaves established pairing records untouched.
```
arturpragacz: >The client generates its Pairing PSK from a CSPRNG and persists it across reboots. 

Is this about the initial one? It could also be provisioned by the manufacturer. In fact that's the more likely case.

>The client MAY rotate it at any time

I don't think clients should be allowed to rotate it themselves.

pairing.md
```diff
@@ -202,7 +216,7 @@ PIN-pairing brute-force protection is built around a per-method failure counter
 The following rules are mandatory for clients implementing `static_pin` or `dynamic_pin`:
 
 - **Per-method failure counter.** The client maintains a failure counter for each PIN-pairing method family (`static_pin` and `dynamic_pin` tracked independently). The counter is persisted across reboots. It is not partitioned by `server_id` or source IP: a single per-method counter for the device.
-- **Increment.** The counter for a method increments on each inner-authentication failure observed in that method's flow.
+- **Increment.** The counter for a method increments on each inner-authentication failure observed in that method's flow: the client's own verification of `server_kc` fails, or [`pair/abort`](#client--server-pairabort) with reason `pin_mismatch` arrives. No other event increments the counter.
```
arturpragacz: We should not increment on received failures, only on our own.

pairing.md
```diff
@@ -286,8 +302,8 @@ Aborts a pairing attempt, started or not. With reason `concurrent_attempt` the s
 - `reason`: string - one of:
   - `attempt_timeout` (client) - the pairing attempt did not complete within the [attempt timeout](#entering-and-leaving-pairing)
   - `concurrent_attempt` (client) - another pairing attempt is already in progress with this client
-  - `locked_out` (client) - the client is in [terminal lockout](#pin-pairing-lockout) for the selected pairing method
+  - `locked_out` (client) - the client is in [terminal lockout#pin-pairing-lockout) for the selected pairing method
```
arturpragacz: Was this change intentional?

pairing.md
```diff
@@ -286,8 +302,8 @@ Aborts a pairing attempt, started or not. With reason `concurrent_attempt` the s
...
   - `method_not_supported` (client) - the server's activity set and `selected_pair_method` are not a permitted combination for the matched PSK, or `selected_pair_method` names a method the client does not currently offer
   - `pin_length_unacceptable` (client) - the `pin_length` in [`server/pair-init`](#server--client-serverpair-init) is below the client's `min_pin_length` or outside the 4–12 range
   - `pin_mismatch` (client or server) - PAKE key-confirmation failed, or (in dynamic PIN pairing) the PIN binding check failed
-  - `user_cancelled` (client or server) - operator aborted the pairing
+  - `user_cancelled` (client) - operator aborted the pairing through a local UI
```
arturpragacz: Why did you remove `or server` here?

arturpragacz (changes requested):

pairing.md
maximmaxim345: I dropped it since this PR also introduces the cancelling `server/activate` ("A server MAY send such a cancelling `server/activate` at any point during a pairing attempt", deliberately not touching the lockout counter), so the server has its own cancel path and `user_cancelled (server)` would be a second overlapping way to do the same thing. Happy to restore it if you see a case the `server/activate` route doesn't cover.

pairing.md
maximmaxim345: Huh, somehow missed that typo...

messaging.md
maximmaxim345: You're right. Thanks!

pairing.md
arturpragacz: I think what's nice about `pair/abort` is that it states the reason why it happened. So if the client has for example some UI it may want to display it there?

maximmaxim345: Sounds good to me too, restored `user_cancelled` for servers. Also added that a server cancelling on operator action SHOULD send it before the cancelling `server/activate`.

arturpragacz (approved):

PR #124, opened 2026-07-21
Reorganize the single-page section order

maximmaxim345:
Builds on #114. This one only moves sections around, no content changes:

- The overview sequence diagram comes out of the "Suggested correction strategy" subsection and becomes its own "Protocol overview" section near the top.
- "Playback Synchronization" moves down next to the player role's audio sections.

Both are plain moves, so the `README.md` diff is just those two blocks changing place. #114 and this PR together produce the same `README.md` as the original combined change.

Depends on #114.


kahrendt (approved):

kahrendt:
Thanks!

PR #125, opened 2026-07-22
Standardize the pairing token format

maximmaxim345:
Defines the **pairing token**: the single `SP:`-prefixed, versioned, base32 string that distributes a client's static key and its Sendspin Pairing PSK together for the Pairing PSK flow. This standardizes the format that [aiosendspin's `pairing_token.py`](https://github.com/Sendspin/aiosendspin/blob/755f34b804ad776eb702c31caf5d3527db0c182e/aiosendspin/noise/pairing_token.py) already produces and consumes, so tokens interoperate across implementations.

Also, I defined the scheme for the QR code, which we previously mentioned but never elaborated: it's the token string carried verbatim, with no URI wrapper, so a scan and a copy/paste yield identical input.

The wire format matches aiosendspin exactly. What the spec adds on top:
- Case-insensitive, lenient decoding is now a normative requirement (trim, upper-case, then strip the optional `SP:` prefix). While aiosendspin already behaves this way, this requires it.
- A `MUST` that the server confirm the decoded `client_key` matches the connection's `client_id` before pairing, which the library doesn't state.


README.md
```diff
@@ -732,6 +732,44 @@ sequenceDiagram
 
 If a Sentinel-keyed connection is already open when the operator picks `pairing_psk`, the server first [re-handshakes](#re-handshake) to the Pairing PSK before sending the `server/activate` shown above.
 
+#### Pairing Token
+
+A server needs both the [Sendspin Pairing PSK](#definitions) and the client's static public key to select and verify the client's Noise identity. The two are distributed together as a **pairing token**: a single case-insensitive ASCII string the operator transfers out of band (copy/paste, QR scan, read aloud) into the server to begin the [Pairing PSK Flow](#pairing-psk-flow). A client offering `pairing_psk` SHOULD surface the token rather than the bare PSK.
```
arturpragacz: > transfers out of band (copy/paste, QR scan, read aloud)

Read aloud? :D

Let's not even suggest that.

README.md
```diff
@@ -732,6 +732,44 @@ sequenceDiagram
...
+
+A token is a fixed `SP:` prefix, a version, and a base32-encoded body:
+
+```
+token   = "SP:" || version || body
+payload = client_key (32 bytes) || pairing_psk (32 bytes)
+```
+
+- `version` - a single decimal digit. This document defines version `1`.
```
arturpragacz:
```suggestion
- `version` - a single alphanumeric character. This document defines version `1`.
```
We can have up to 26+9 versions with letters following the digits.
Also I could swear I made the first version `0`, but no, it is actually `1`. Maybe we should change it while there's still time?

arturpragacz (changes requested):

README.md
maximmaxim345: Also pushed a commit to change it to 0 here in the Spec.

README.md
maximmaxim345: Haha, that would be a good contender for r/badUIbattles though 😂.
Dropped it so nobody actually tries that.

arturpragacz (approved):

PR #126, opened 2026-07-23
Add announcement role (announcement@v1)

marcelveldt:
Adds a new optional role, `announcement@v1`, that lets a client play short client-specific audio (voice-assistant replies, doorbell chimes, alerts) as its own stream, mixed over or independent of media playback, with the media ducked locally on the client.

## Why

Announcements are currently outside the protocol, so servers fake them. Music Assistant today snapshots player state, ungroups, stops the stream, plays the clip as normal media, then restores everything. For ESPHome speakers that is a step backwards: their firmware already has a first-class dual-pipeline model (media keeps playing, the announcement mixes over it, the media ducks) that Sendspin can't reach.

This role announces on any speaker without touching a playing group's timeline. The media stream keeps flowing to every group member, and each targeted client ducks its own output and mixes the announcement over it.

## What's in it

- A client advertises the role plus an `announcement@v1_support` object (its announcement formats and a dedicated buffer capacity).
- The server starts a per-client stream with `stream/start` carrying the codec params, a single `start_timestamp`, the duck level and ramp, and an optional absolute `volume`.
- Audio is pushed as a new binary type 24 (role 6, IDs 24-27). Chunks are untimed and played continuously from `start_timestamp`; if TTS is slow the stream simply stretches. `stream/end` finishes or aborts it.
- Ducking is bound to the stream's lifetime, so a dropped connection can never leave the media ducked.
- Multi-speaker is one stream per targeted client with a shared `start_timestamp` for a coordinated start. Sample-accurate cross-client sync of the announcement audio itself is out of scope for v1.
- Touches `messaging.md` and `template.md` for the role registration, binary-ID allocation, and message objects; adds `roles/announcement/v1.md`.

Landing directly as `announcement@v1` (pre-1.0 spec), not staged as a draft revision.

## Out of scope (possible follow-up)

- **Second-server announcements.** An announcement rides the client's single admitted connection, so the server streaming music also handles announcements. Home Assistant routes through Music Assistant for this today. Letting a second server come in, announce, and leave while another streams music would need announcement-capable secondary connections in the connection admission model. Left out for now: real added complexity for little gain. A follow-up if demand shows up.

## Open questions

- **Duck defaults.** `media_duck_db` defaults to 0 (off) and `duck_ramp_ms` to 100 ms. Input from the ESPHome side welcome.

Companion client reference implementation: Sendspin/sendspin-cpp#101.


balloob:
> Announcements are strictly per-client. A multi-room announcement is N independent streams; cross-client announcement sync is an explicit non-goal for v1.

Doesn't this defeat the purpose? The nice thing about MA doing the announcement is that all speakers in my living room mute at the same time, play the announcement, and then continue playing. If only 1 player ducks the music, I might still miss it?

Also, this proposal made me initially think that announcements would be possible to a second connected client, that is not the one currently streaming the music (as Sonos allows), but that doesn't seem to be the case. 

All-in-all, I think this proposal could use some more thought about the use cases we want to unlock and if the current proposal covers these.

marcelveldt:
1) Announcements are always **per speaker**, not per grouped speakers.
2) They can optionally be addressed to multiple speakers at once but that should be user choice, not a protocol limitation.
3) It can be nice to also support coordinated playback of an announcement to multiple playersat once but that is more a nice to have than a must-have.
4) The player itself is responsible for mixing in an announcement stream. It announces this role only if it has this capability.

Requirements 1, 2, 3 come from experience with this after having this feature in MA for a few years.
It is used to announce something in a specific room / speaker more often than everywhere.
If users want to announce something at multiple speakers at once, they usually just target multiple speakers at once with the same announcement (and MA starts them all at the same time) or they create a dedicated group for it.

> Also, this proposal made me initially think that announcements would be possible to a second connected client, that is not the one currently streaming the music

Yes that is exactly how it should work - is that not clearly described enough ?
This way you can also unlock HA streaming announcements over the sendspin route

EDIT: tightened the wording a bit as I can see where the confusion came from.

balloob:
But if it's per speaker, it won't work with stereo pairs, or any other set up where a user has 2 speakers set up in the same room. Or if a user targets an area in Home Assistant, it would end up hearing the same announcement in 1 room slightly out of sync. Or you think that HA could be smart to see it's targeting 2 Sendspin players, and get a sync lock before playing the announcement without too much delay?

marcelveldt:
> But if it's per speaker, it won't work with stereo pairs, or any other set up where a user has 2 speakers set up in the same room. Or if a user targets an area in Home Assistant, it would end up hearing the same announcement in 1 room slightly out of sync. Or you think that HA could be smart to see it's targeting 2 Sendspin players, and get a sync lock before playing the announcement without too much delay?

It has coordinated start, which is perfect for this usecase

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+- Announcement audio MUST NOT be reflected in visualizer or color output, and does not alter metadata or artwork state - those roles describe the media stream only.
+- There is no `stream/request-format` for the announcement role: the server picks a fresh format from the client's `supported_formats` for every announcement.
+
+### Server behavior for announcements
+
+- The server MUST NOT send any announcement-scoped message (`stream/start` with an `announcement` object, `stream/clear`/`stream/end` listing `announcement`, or type `24` binary) to a client whose `active_roles` do not include the announcement role.
+- The chunk duration bounds of the player role apply (15-150 ms; see [Server Audio Send Constraints](../player/v1.md#server-audio-send-constraints)).
+- The first chunk's timestamp SHOULD be at least the client's announcement `required_lead_time_ms` (or the 500 ms default) past the `stream/start` server transmit time.
```
maximmaxim345, line 101: This ignores `duck_ramp_ms`

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+  - `duck_ramp_ms?`: integer - ramp duration in milliseconds (range 0-2000, default 100) for both applying and releasing the ducking
+  - `volume?`: integer - range 0-100. When present, the client SHOULD render the announcement at the loudness that master volume `volume` would produce, regardless of the current master volume (see [Mixing and Ducking](#mixing-and-ducking)). When absent, the announcement follows the current master volume
+
+The format MUST be one the client listed in its [`supported_formats`](#client--server-clienthello-announcementv1-support-object).
+
+Re-sending `stream/start` with an `announcement` object while an announcement stream is active updates the configuration without clearing buffers (e.g., to change the duck level mid-announcement); a new `media_duck_db` target ramps from the current gain. To replace the announcement audio itself, the server MUST first send [`stream/clear`](../../messaging.md#server--client-streamclear) listing the `announcement` role.
+
+### Server → Client: `stream/clear` announcement
```
maximmaxim345, line 50: lets drop this, don't see a use case where `stream/clear` should be used instead of a `stream/end`.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+### Server → Client: `stream/end` announcement
+
+When [`stream/end`](../../messaging.md#server--client-streamend) lists the `announcement` role (or omits `roles`, which ends all active streams), the announcement stream is over: clients stop announcement output, discard any remaining buffered announcement audio, release the ducking (ramped over `duck_ramp_ms`), and report `state: 'idle'` if they report announcement state.
+
+### Server → Client: Announcement Chunks (Binary)
+
+Binary messages SHOULD be rejected if there is no active announcement stream or the client is not [`available`](../../messaging.md#client--server-clientstate).
+
+- Byte 0: message type `24` (uint8)
+- Bytes 1-8: timestamp (big-endian int64) - server clock time in microseconds when the first sample should be output
```
maximmaxim345, line 63: We can drop the timestamp here and instead have a single timestamp in `stream/start`.
The client will just play them continuously and it's easier to implement the case where it arrived too late (since you just play them continuously).
And it will still start in sync with a best effort mechanism. Not having `stream/clear` support also closes edge cases.

And that would also implicitly solve the slow TTS issues: the client will try to play them continuously, but if its coming it too slow it will automatically be a "best effort" approach.

`buffer_capacity` needs some more thought though.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Mixing and Ducking
+
+A client with both an active media stream and an active announcement stream outputs both concurrently, mixed locally. The gain stages are ordered as follows:
+
+1. `media_duck_db` applies to the media signal only, before mixing.
+2. The announcement `volume` (when present) applies to the announcement signal only, before mixing. Since master volume is applied after the mix, the client compensates the announcement gain so the announcement is rendered at the loudness that master volume `volume` would produce. Implementations MAY snapshot the master volume at announcement start for this compensation.
+3. The master volume and mute of the [player role](../player/v1.md#client--server-clientstate-player-object) apply to the mixed output. Announcements are therefore silenced on a muted client.
+
+Starting an announcement stream MUST NOT pause, stop, mute, or shift the timeline of the client's media stream - ducking is a gain change only. This keeps a grouped client's media playback sample-identical with its group while it plays an announcement.
```
maximmaxim345, line 80: Nit: it was never sample-identical due to the small corrections to keep it in sync.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+- The chunk duration bounds of the player role apply (15-150 ms; see [Server Audio Send Constraints](../player/v1.md#server-audio-send-constraints)).
+- The first chunk's timestamp SHOULD be at least the client's announcement `required_lead_time_ms` (or the 500 ms default) past the `stream/start` server transmit time.
+- `buffer_capacity` is a hard per-client byte limit on outstanding un-played compressed announcement audio.
+- The server knows when the announcement has finished playing from its own timeline (the last chunk's end timestamp). For normal completion it SHOULD NOT send `stream/end` for the announcement role before that time has passed; sending it earlier is the abort path.
+- Announcements ride the client's admitted connection: the server allowed to stream media may announce. Letting a *second* server (e.g. a voice assistant) announce while another server streams media requires announcement-capable secondary connections in the [connection admission model](../../connection.md#multiple-servers-server-initiated) - a companion change tracked separately from this role.
```
maximmaxim345, line 104: This should belong in a PR description and/or issue instead of the spec IMO.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+
+The `announcement` object in [`stream/start`](../../messaging.md#server--client-streamstart) has this structure:
+
+- `announcement`: object
+  - `codec`: string - codec to be used
+  - `sample_rate`: integer - sample rate to be used
+  - `channels`: integer - channels to be used
+  - `bit_depth`: integer - bit depth to be used
+  - `codec_header?`: string - codec header encoded as standard Base64, if necessary (e.g., FLAC)
+  - `media_duck_db?`: integer - reduction in decibel (range 0-50, default 0) to apply to this client's own media output while the announcement stream is active. 0 means no ducking
```
maximmaxim345, line 42: Mention what the db is relative to (amplitude, 20*log10), the player role is explicit about its scale too.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+- The first chunk's timestamp SHOULD be at least the client's announcement `required_lead_time_ms` (or the 500 ms default) past the `stream/start` server transmit time.
+- `buffer_capacity` is a hard per-client byte limit on outstanding un-played compressed announcement audio.
+- The server knows when the announcement has finished playing from its own timeline (the last chunk's end timestamp). For normal completion it SHOULD NOT send `stream/end` for the announcement role before that time has passed; sending it earlier is the abort path.
+- Announcements ride the client's admitted connection: the server allowed to stream media may announce. Letting a *second* server (e.g. a voice assistant) announce while another server streams media requires announcement-capable secondary connections in the [connection admission model](../../connection.md#multiple-servers-server-initiated) - a companion change tracked separately from this role.
+- There is no negative acknowledgement; a client MAY silently ignore an announcement stream it cannot currently service.
```
maximmaxim345, line 105: this is client behavior, not server. move it to the lifecycle section.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+### Interaction with Media, Groups, and Other Roles
+
+- Announcement streams are independent of group membership and of the media stream lifecycle: group changes, media `stream/start`, media-scoped `stream/clear`, and media `stream/end` do not affect an active announcement stream (ducking simply becomes a no-op when media ends). [`group/update`](../../messaging.md#server--client-groupupdate) describes the media group only.
+- Announcement audio MUST NOT be reflected in visualizer or color output, and does not alter metadata or artwork state - those roles describe the media stream only.
+- There is no `stream/request-format` for the announcement role: the server picks a fresh format from the client's `supported_formats` for every announcement.
+
+### Server behavior for announcements
+
+- The server MUST NOT send any announcement-scoped message (`stream/start` with an `announcement` object, `stream/clear`/`stream/end` listing `announcement`, or type `24` binary) to a client whose `active_roles` do not include the announcement role.
+- The chunk duration bounds of the player role apply (15-150 ms; see [Server Audio Send Constraints](../player/v1.md#server-audio-send-constraints)).
```
maximmaxim345, line 100: We try to limit cross role links since roles should be as self contained as possible.
Inline the 15-150ms here, same for `required_lead_time_ms`.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+
+When [`stream/end`](../../messaging.md#server--client-streamend) lists the `announcement` role (or omits `roles`, which ends all active streams), the announcement stream is over: clients stop announcement output, discard any remaining buffered announcement audio, release the ducking (ramped over `duck_ramp_ms`), and report `state: 'idle'` if they report announcement state.
+
+### Server → Client: Announcement Chunks (Binary)
+
+Binary messages SHOULD be rejected if there is no active announcement stream or the client is not [`available`](../../messaging.md#client--server-clientstate).
+
+- Byte 0: message type `24` (uint8)
+- Bytes 1-8: timestamp (big-endian int64) - server clock time in microseconds when the first sample should be output
+- Rest of bytes: encoded audio frame
```
maximmaxim345, line 64: This doesn't define how data (and PCM) is encoded. See player/v1.md.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Announcement Playback Behavior
+
+### Mixing and Ducking
+
+A client with both an active media stream and an active announcement stream outputs both concurrently, mixed locally. The gain stages are ordered as follows:
+
+1. `media_duck_db` applies to the media signal only, before mixing.
+2. The announcement `volume` (when present) applies to the announcement signal only, before mixing. Since master volume is applied after the mix, the client compensates the announcement gain so the announcement is rendered at the loudness that master volume `volume` would produce. Implementations MAY snapshot the master volume at announcement start for this compensation.
+3. The master volume and mute of the [player role](../player/v1.md#client--server-clientstate-player-object) apply to the mixed output. Announcements are therefore silenced on a muted client.
```
maximmaxim345, line 78: Line 6 says this can be advertised standalone, but then there's no player object and no volume or mute at all.

kepstin: I suspect that being able to override the mute on a media playback stream so the announcement can be heard (i.e. media playback stays muted, only announcement is audible) might be preferred in some cases.

Might need to be optional, so people can decide in automations on a case-by-case basis (e.g. consider time of day, room occupation, etc.)

roles/announcement/v1.md
kepstin: Yeah, re-wording this to say that the reason for this is to keep sync with other players in the group might be the right fix?

```suggestion
Starting an announcement stream MUST NOT pause, stop, mute, or shift the timeline of the client's media stream - ducking is a gain change only. This keeps a grouped client's media playback in sync with its group while it plays an announcement.
```

roles/announcement/v1.md
marcelveldt: good catch - the start timestamp now budgets required_lead_time_ms + duck_ramp_ms, so the duck can ramp in before audio starts.

roles/announcement/v1.md
marcelveldt: dropped it. went with your single start-timestamp idea below, so there's no clear/replace path anymore.

roles/announcement/v1.md
marcelveldt: adopted this. single `start_timestamp` in stream/start now, chunks are untimed and played continuously - slow TTS just stretches the stream (best effort). announcement stream/clear is gone. kept `buffer_capacity` as a plain flow-control byte cap for now; happy to revisit if you want something different there.

roles/announcement/v1.md
marcelveldt: reworded to "in sync with its group" per kepstin's suggestion.

roles/announcement/v1.md
marcelveldt: agreed, moved it out of the spec into the PR description as an open question.

roles/announcement/v1.md
marcelveldt: added the scale - a reduction of x dB scales the media amplitude by 10^(-x/20).

roles/announcement/v1.md
marcelveldt: moved it to the lifecycle section.

roles/announcement/v1.md
marcelveldt: inlined both - the 15-150 ms chunk bound and the required_lead_time_ms definition are spelled out in this file now instead of linking the player role.

roles/announcement/v1.md
marcelveldt: added the frame encoding: pcm is little-endian two's complement, interleaved by channel (24-bit packed as 3 bytes); opus/flac carry one encoded frame per message.

roles/announcement/v1.md
marcelveldt: good point. fixed the standalone gap: master volume/mute come from the player object when that role is present, otherwise the client applies its own local volume/mute. and added an optional `override_mute` flag (default false) for the alarm case @kepstin raised - media stays muted, the announcement is still heard.

marcelveldt:
Reworked and rebased on main. Main changes since your review:

- Adopted @maximmaxim345's single start-timestamp idea: one `start_timestamp` in `stream/start`, chunks are untimed and played continuously (slow TTS just stretches, best effort). Dropped the announcement `stream/clear` entirely - replacing a clip is now `stream/end` + a new `stream/start`.
- Added an optional `override_mute` flag (default false) for the alarm case @kepstin raised - media stays muted, the announcement is still heard.
- Addressed the smaller points too: dB/amplitude scale on `media_duck_db`, pcm/opus/flac frame encoding spelled out, inlined the cross-role bits (15-150ms, `required_lead_time_ms`), folded `duck_ramp_ms` into the start lead-time budget, and closed the standalone volume/mute gap.
- Caught an Opus mandate that had slipped back in and fixed it to mirror the player role (servers MUST support flac+pcm, opus optional, per #238).
- Second-server announcing moved out of the spec into an open question in the description.

Reference client implementation updated to match in Sendspin/sendspin-cpp#101. Marking ready for another look.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Client → Server: `client/state` announcement object
+
+The `announcement` object in [`client/state`](../../messaging.md#client--server-clientstate) informs the server of announcement-specific state. Only for clients with the `announcement` role. All fields are optional and purely informational - the server derives announcement completion from the stream timeline regardless (see [Server Behavior for Announcements](#server-behavior-for-announcements)) - so a minimal client MAY never send this object.
+
+- `announcement`: object
+  - `state?`: 'playing' | 'idle' - whether announcement audio is currently being output. SHOULD be sent on transitions
+  - `required_lead_time_ms?`: non-negative integer - minimum startup lead time in milliseconds for the announcement pipeline (codec init, decode warmup, mixer attach), measured from the server transmit time of [`stream/start`](#server--client-streamstart-announcement-object) to the announcement's `start_timestamp`. When absent, the server SHOULD assume a conservative default (500 ms is recommended)
```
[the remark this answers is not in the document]
marcelveldt: fair - the role does define a state object, so per messaging.md the client must send it on activation. reworded: the client MUST include the announcement object when the role is activated (an empty `{}` if it has nothing to report), fields stay optional.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+  - `media_duck_db?`: integer - reduction applied to this client's own media output while the announcement stream is active, in decibels of linear amplitude (a reduction of `x` dB scales media amplitude by `10^(-x/20)`). Range 0-50, default 0; 0 means no ducking
+  - `duck_ramp_ms?`: integer - ramp duration in milliseconds (range 0-2000, default 100) for both applying and releasing the ducking
+  - `volume?`: integer - range 0-100. When present, the client SHOULD render the announcement at the loudness that master volume `volume` would produce, regardless of the current master volume (see [Mixing and Ducking](#mixing-and-ducking)). When absent, the announcement follows the current master volume
+  - `override_mute?`: boolean - default false. When true, a muted client still renders the announcement while its media stays muted (see [Mixing and Ducking](#mixing-and-ducking))
```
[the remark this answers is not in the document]
marcelveldt: good catch on the master-volume-at-0 case. reworked the mixing model so a present `volume` is an absolute level that bypasses the master gain entirely - the master volume is not applied to it, so it holds its loudness even at master 0. the override_mute path is now defined per-signal too.

roles/announcement/v1.md
```diff
@@ -0,0 +1,127 @@
...
+- Announcement audio MUST NOT be reflected in visualizer or color output, and does not alter metadata or artwork state - those roles describe the media stream only.
+- There is no `stream/request-format` for the announcement role: the server picks a fresh format from the client's `supported_formats` for every announcement.
+
+### Server Behavior for Announcements
+
+- The server MUST NOT send any announcement-scoped message (`stream/start` with an `announcement` object, `stream/end` listing `announcement`, or type `24` binary) to a client whose `active_roles` do not include the announcement role.
+- Each announcement chunk SHOULD carry between 15 ms and 150 ms of audio.
+- The announcement's `start_timestamp` SHOULD be at least the client's announcement `required_lead_time_ms` (or the 500 ms default) plus `duck_ramp_ms` past the `stream/start` server transmit time, so the client can warm up its pipeline and ramp the duck in before output begins.
+- `buffer_capacity` is a hard per-client byte limit on outstanding un-played compressed announcement audio.
```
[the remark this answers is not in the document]
marcelveldt: reworded - it's a best-effort flow-control bound now, not a hard acked cap. the server keeps outstanding audio within `buffer_capacity` by estimating consumption from the nominal timeline (start + elapsed at the sample rate); there's no per-chunk ack.

roles/announcement/v1.md
```diff
@@ -0,0 +1,130 @@
...
+## Announcement Playback Behavior
+
+### Mixing and Ducking
+
+A client with both an active media stream and an active announcement stream mixes the two locally. Gain applies per signal:
+
+- **Media:** `media_duck_db` reduces the media signal before mixing; the client's master volume then applies to it, and mute silences it.
+- **Announcement without `volume`:** follows the client's master volume like the media, and mute silences it.
+- **Announcement with `volume`:** rendered at the fixed linear amplitude that master volume `volume` would produce, with the master volume NOT applied to it, so it holds that loudness regardless of the current master volume or of media ducking. Mute still silences it. Rendering it at a fixed level independent of the post-mix master gain is what lets `volume` deliver its loudness even when the master volume is low or zero.
+- **`override_mute`:** when the stream sets it, mute does not silence the announcement (the media stays muted); the announcement plays at its `volume`, or at the current master volume when `volume` is absent.
```
kahrendt: This is difficult to do on devices where the mute is handled outside the Sendspin client.

If a client has a hardware mute, we could potentially mute the media stream in software and then unmute at the hardware level. A client could adjust the required lead time to try and account for this, but it's still a guessing game. I have no idea how fast the hardware unmute will get applied. In ESPHome, for example, this has to work its way through the main loop centered defers and callbacks, which are very variable. Getting it wrong means a sudden loud burst of music or having to play it extremely conservative in the required lead time.

roles/announcement/v1.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `announcement`: object
+  - `codec`: 'opus' | 'flac' | 'pcm' - codec to be used
+  - `sample_rate`: integer - sample rate to be used
+  - `channels`: integer - channels to be used
+  - `bit_depth`: integer - bit depth to be used; ignored for `opus`
+  - `codec_header?`: string - codec header encoded as standard Base64, if necessary (e.g., FLAC)
+  - `start_timestamp`: integer - server clock time in microseconds when announcement output should begin. The client translates it via [clock synchronization](../../messaging.md#clock-synchronization) and starts at that time, or as soon as possible if it has already passed. Targeting several clients with the same `start_timestamp` gives a coordinated multi-speaker start. A client that has not established clock synchronization (e.g. an announcement-only client, which is not required to) starts as soon as possible and does not take part in a coordinated start
+  - `media_duck_db?`: integer - reduction applied to this client's own media output while the announcement stream is active, in decibels of linear amplitude (a reduction of `x` dB scales media amplitude by `10^(-x/20)`). Range 0-50, default 0; 0 means no ducking
```
kahrendt: If you set the lower bound range because of the current ESPHome restriction, that is not necessary. It's a software only limit that I need to remove in a PR; underneath the hood it works to -infinity dB; i.e., total silence. We should support total silence for the media using a very large dB reduction value like 200 dB, which is more than enough for silence on audio with 32-bits per sample

roles/announcement/v1.md
```diff
@@ -0,0 +1,130 @@
...
+  - `duck_ramp_ms?`: integer - ramp duration in milliseconds (range 0-2000, default 100) for both applying and releasing the ducking
+  - `volume?`: integer - range 0-100. When present, the announcement is rendered at the fixed loudness that master volume `volume` would produce, independent of the current master volume (see [Mixing and Ducking](#mixing-and-ducking)). When absent, the announcement follows the current master volume
+  - `override_mute?`: boolean - default false. When true, a muted client still renders the announcement while its media stays muted (see [Mixing and Ducking](#mixing-and-ducking))
+
+The format MUST be one the client listed in its [`supported_formats`](#client--server-clienthello-announcementv1-support-object).
+
+Re-sending `stream/start` with an `announcement` object while an announcement stream is active updates its configuration (e.g., the duck level or volume) without clearing buffers; a new `media_duck_db` target ramps from the current gain. To play a different clip, the server first ends the current announcement with [`stream/end`](#server--client-streamend-announcement) and then starts a new one.
```
kahrendt: What happens if `duck_ramp_ms` changes while an announcement stream is active?

roles/announcement/v1.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Server → Client: `stream/end` announcement
+
+An announcement stream ends only when [`stream/end`](../../messaging.md#server--client-streamend) lists the `announcement` role explicitly; a `stream/end` that omits `roles` (a media stop) leaves an active announcement running. When `announcement` is listed, no more announcement audio follows, so the client plays out any announcement audio still buffered, then releases the ducking (ramped over `duck_ramp_ms`) and reports `state: 'idle'` if it reports announcement state. The server sends this once it has delivered the final chunk, which is the normal end: because the client drains its buffer before releasing, a normal end never truncates the announcement. Sending it before all audio is delivered stops the announcement early - the client still plays out what it has buffered, and only the audio the server had not yet sent is lost. This early stop is the closest thing to an abort; there is no separate instant cancel in v1.
```
kahrendt: This paragraph especially doesn't fit the tone/style of the spec, especially the "no separate instant cancel in v1"

roles/announcement/v1.md
```diff
@@ -0,0 +1,130 @@
...
+
+The stream carries a single `start_timestamp` (in [`stream/start`](#server--client-streamstart-announcement-object)); the chunks themselves are untimed. The client decodes them in order and plays them continuously, beginning at `start_timestamp`, and MUST NOT drop a chunk merely because it arrived late. If announcement audio is produced slower than real time (e.g., streaming text-to-speech), the stream simply stretches: the client plays each chunk as it arrives, a best-effort schedule rather than a forced timeline. The [player sync accuracy rules](../player/v1.md#playback-synchronization) do not apply to announcement output.
+
+## Announcement Playback Behavior
+
+### Mixing and Ducking
+
+A client with both an active media stream and an active announcement stream mixes the two locally. Gain applies per signal:
+
+- **Media:** `media_duck_db` reduces the media signal before mixing; the client's master volume then applies to it, and mute silences it.
```
kahrendt: Just to re-iterate the other comment, we should use a very large dB reduction to silence the music volume. Mute shouldn't be used to control the level of just one stream.

kahrendt (commented):
In general, the text in the PR doesn't fit the style and tone of the rest of the spec. It should be able to be condensed down quite a bit.

roles/announcement/v1.md
kahrendt: As a follow up, I don't think we should support this as is. The workaround with client delay might work, but there are many variables outside of our control that could cause issues.

marcelveldt: makes sense - dropped `override_mute` entirely. mute stays a global control; to make the media silent under an announcement you use a large `media_duck_db` instead.

roles/announcement/v1.md
marcelveldt: removed the 0-50 cap - `media_duck_db` is a non-negative integer now, default 0, and a large value (e.g. 200) silences the media.

roles/announcement/v1.md
marcelveldt: defined it: a re-sent stream/start updates duck level, ramp and volume in place - a new duck level ramps from the current gain over the new `duck_ramp_ms`, which also applies to the release.

roles/announcement/v1.md
marcelveldt: rewrote it - condensed to the normative behavior and dropped the abort/instant-cancel commentary. did the same pass over the rest of the file.

roles/announcement/v1.md
marcelveldt: agreed, that's the model now: mute is global, silence comes from a large `media_duck_db`. `override_mute` is gone.

marcelveldt:
Did a condensing pass over the whole role file to match the tone of the other roles, and dropped `override_mute` per @kahrendt: mute stays a global control, and the media is silenced under an announcement with a large `media_duck_db` (the 0-50 cap is gone). Also defined what a re-sent `duck_ramp_ms` does. cpp reference is being updated to match.

PR #130, opened 2026-08-04
Improve PIN pairing window

arturpragacz:
Improve PIN pairing window.

Unifies PIN-pairing consent around the pairing window and replaces terminal lockout with escalation.

- **Pairing window as the single consent primitive.** The window (previously static-PIN-only) now gates any *gesture-gated* attempt: every `static_pin` attempt, and `dynamic_pin` attempts when the method is escalated or the session's PIN is shorter than 6 digits.
- **Lockout → escalation.** The failure counter (now `dynamic_pin`-only) no longer ends in terminal lockout: at 10 failures the method is *escalated* — every attempt becomes gesture-gated until a successful round resets it. The `locked_out` field (descriptor & config) and abort reason are gone; `get-pairing-config` reports `escalated` instead. Static PIN needs no counter: each attempt already costs a window.
- **New `client/pair-pending`.** Signals that an attempt is gesture-gated and awaiting a window, without starting the attempt; index-gated like `client/pair-init`.
- **New `management/open-pairing-window`.** Lets a paired server open the window in place of the operator gesture.
- **`server/activate` shape.** `selected_pair_method` becomes a `pairing` object; `pin_length` moves from `server/pair-init` into the activation so the client can validate it and decide gating before anything starts.
- **Method advertisement & provisioning.** `supported_pair_methods` is clarified to list only currently-offered (enabled) methods. Out-channel devices SHOULD ship `static_pin` too — disabled, with no PIN provisioned; enabling it without a configured (or simultaneously supplied) PIN is rejected.


maximmaxim345 (approved):
Nice, changes look good to me.
Really nice to not have a lockout anymore!

PR #131, opened 2026-08-04
Add language hint for spoken dynamic PIN

arturpragacz:
Add language hint for spoken dynamic PIN.


messaging.md
```diff
@@ -210,6 +210,7 @@ Only after receiving the initial `server/activate` should the client send any ot
 - `pairing?`: object - parameters of the pairing attempt this activation admits. Required when `'pairing'` is in `activities`; absent otherwise. A client ignores this field when `activities` does not include `'pairing'`.
   - `method`: 'dynamic_pin' | 'pairing_psk' | 'static_pin' - pairing method the server picked, drawn from the client's `supported_pair_methods`.
   - `pin_length?`: integer - the dynamic [PIN length](pairing.md#dynamic-pin-pairing-flow) for this session. Required when `method` is `'dynamic_pin'`; absent otherwise.
+  - `languages?`: string[] - non-empty list of [BCP 47](https://www.rfc-editor.org/rfc/rfc5646) language tags in descending operator preference (e.g. `["ca", "es", "en"]`), for spoken [PIN emission](pairing.md#dynamic-pin-pairing-flow). Optional when `method` is `'dynamic_pin'`; absent otherwise.
```
maximmaxim345, line 213: Just an idea: should we allow sending `languages` regardless of the selected method? Still optional for the server, but clients could use it for other things, like translating the UI they show.

maximmaxim345 (approved):
Looks good, just left one comment about possibly expanding `languages` for other uses too.

But thats not blocking this PR and just an idea. We can discuss/add that separately to this change.

messaging.md
arturpragacz: Hmm, if we would want a hint for translating the whole client UI, then I think we would have to make it a global option, and not just here in pairing then.

Issue #128, opened 2026-07-30
Artwork dimensions letterboxing ambiguity


kahrendt:
aiosendspin letterboxes artwork: https://github.com/Sendspin/aiosendspin/blob/9212f920e8fbaf9ad357b43835bd32cc386e73b8/aiosendspin/server/roles/artwork/group.py#L161-L181

The spec says "Note: The server will scale images to fit within the specified dimensions while preserving aspect ratio."

By my read, that makes it seem like it won't letterbox, but instead not exceed the dimensions.

My preference is the server letterboxes, it makes support on the client side easier.

maximmaxim345:
I actually haven't noticed that this could also be interpreted in a different way but you're right. We should require servers to letterbox.

Should we also define that padding needs to be black?

maximmaxim345:
Was thinking a bit more about this. Servers should letterbox, so the image arrives at exactly the dimensions declared for the channel, but we should not mandate how the padding looks. Mandating is simple, but I can imagine better ways to fill that space, like a blurred version of the cover in the background, or just using an extracted color of the artwork.

One thing I would pin down is that the image is uncropped, otherwise part of it might get cut off.

Was also thinking about whether we want to allow clients to ask for a not letterboxed version, but that is IMO outside the scope here and something for later/`artwork@v2`, since it needs more thought (especially with #135).

kahrendt:
I agree with not mandating how the padding should look, but I also imagine using an extracted color could really clash with how the client is configured; e.g., it has its own color gradient fore surrounding background.

Perhaps it would be better to drop the letterbox, I can make that work on the client side. Then clients can determine how they should fill in the space instead.

balloob:
We are going to letterbox, black background. We're going to update phrasing for this.

maximmaxim345:
We need to call `media_width` just `width` (and `height` as well)

PR #135, opened 2026-08-13
Define scheduled updates for metadata, color, and artwork

maximmaxim345:
Define what a future timestamp on `metadata`, `color`, and artwork messages means: the update is scheduled and takes effect at that time, with one shared current-plus-pending model where the newest message always wins the pending slot.

## Changes

The `metadata` and `color` state timestamps left the future-timestamp case undefined, and the artwork binary rule ("latest wins") could be read as either arrival order or timestamp order, so client behavior diverged and a server could not safely send an update ahead of time (e.g. the next track's metadata timed to the audible track change, which lags the server's own transition by the buffered audio).

All three roles now define the same model. Clients keep their current state plus at most one pending update: a future-timestamped message becomes the pending update, replacing any held one, and applies when its time is reached; any other message applies immediately and discards the held one. Clients never compare timestamps. Around it, `server/state` defers to the role sections instead of implying apply-on-receipt, a `null` role object discards any pending update, omitting a role object leaves a pending update in place, the first state message for a role must carry a past-or-present timestamp, progress extrapolation reads only from the current state and never from a pending update, an empty artwork message is a scheduled clear, and `stream/end` discards pending images. There is no wire change and no role version bump; this pins down what v1 left undefined.

This solves the problem without needing `stream/clear`. `stream/clear` would not properly solve this problem, since:
- It is only defined on a stream basis. `metadata` and `color` don't use binary streams.
- `stream/clear` would clear all pending artwork channels, this would not look great in practice, causing disappearing images.

## What #175 removed

Rebased onto main now that #175 has landed. Dropping delta merging retired the rule this PR previously needed: that the first message sent after a scheduled update repeat every field the scheduled message carried. With full state on every message, a client that applied the pending update and one that discarded it both land in the next message's state, so there is nothing left to reconcile. The cancel recipe survives in simplified form and now reads the same way in all three roles.

#175 also anchors `track_progress` to the metadata `timestamp`, which is what makes a scheduled `progress` well defined: it describes the position at the moment the update takes effect.

## Changed after review

The lead-time bound now measures from the timestamp rather than from the moment of scheduling: "SHOULD NOT send a scheduled update more than 20 seconds before its `timestamp`". That is a different rule from the original "SHOULD NOT schedule an update more than 20 seconds ahead", and the better one, since it bounds how long a client holds a pending update and therefore how much a pause discards.

## Breaking changes: server

In practice none: AFAIK no server sends timestamps in advance, and timestamping every update with the current time remains fully conformant with no obligations. A server that schedules ahead cancels by resending the current state with a past or present timestamp; artwork needs nothing special since every message is a complete image.

## Breaking changes: client

Clients that apply `metadata`/`color` on receipt render scheduled updates too early once servers start scheduling; they need the single pending slot: hold a future-timestamped message until its time, let any newer message replace it, no timestamp comparison anywhere. Artwork clients that queue multiple timestamped images or block on them must reduce to the same single pending image per channel and discard pending images on `stream/end`.


kahrendt (commented):
Should we specify a suggestion in the spec on how much lead time the server should future data? Doing it early on means potentially more network traffic around the artwork role; i.e., pausing needs to clear the pending buffer and then it would get resent as soon as its resumed. If it only sends it a small time in advance this window gets narrower.

maximmaxim345:
> Should we specify a suggestion in the spec on how much lead time the server should future data? Doing it early on means potentially more network traffic around the artwork role; i.e., pausing needs to clear the pending buffer and then it would get resent as soon as its resumed. If it only sends it a small time in advance this window gets narrower.

My Music Assistant implementation/prototype sends it around 30s in advance, would limiting it to maybe 20s or 15s (with a SHOULD) be fine with you? Or keep it 30s and add a note in the Spec?

It's not explicitly mentioned anywhere, but a nice feature this can allow is a "coming up" display, already showing the next track with a countdown.

For reference, Roon has something similar in their player:
image
(https://www.fidelity-online.de/the-new-roon-radio/)

maximmaxim345:
For now I set the limit at 20 seconds, but I can still change it if you think that isn't optimal.

Also explicitly mentioned how clients might use the data to encourage servers to actually send it in advance.
But I still didn't require servers to send it in advance since that heavily depends on the servers architecture and content.

maximmaxim345:
Found a gap in this PR while I was thinking about #128. beebf5f now discards the pending image if `stream/start` reconfigures that artwork channel.
Alternative would be to keep it and treat a channel's config as describing only the images sent while it is current, but then a channel switched to `source: 'none'` would still display an image seconds after being disabled.

roles/artwork/v1.md
```diff
@@ -66,6 +66,12 @@ The message type determines which artwork channel this image is for:
...
 
-**Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty binary message (only the message type byte and timestamp, with no image data) for that channel.
+**Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty binary message (only the message type byte and timestamp, with no image data) for that channel. An empty message follows the same rules as any other image: a future timestamp schedules the clear.
+
+#### Server rules for scheduled artwork
+
+Servers SHOULD NOT schedule an image more than 20 seconds ahead.
```
kahrendt: "SHOULD NOT send a scheduled update more than 20 seconds before its timestamp" is less ambiguous.

roles/color/v1.md
```diff
@@ -6,10 +6,22 @@ This section describes messages specific to clients with the `color` role, which
...
+
+#### Scheduled color updates
+
+A `color` object whose `timestamp` is in the future is a scheduled update: state that takes effect at that time (for example, the next track's colors timed to the audible track change).
+
+Clients keep a **current state** plus at most one **pending update**. The current state is the running merge of applied updates and is what the client renders from. A message whose `timestamp`, translated to the local clock via the [time filter](../../messaging.md#clock-synchronization) (current best estimate, no waiting for convergence), is still in the future becomes the pending update, replacing any held one, and is applied to the current state when that moment is reached. A message whose translated timestamp is in the past or present is applied immediately and discards any held pending update. Clients MAY ease into the pending update around its timestamp (e.g. a color blend).
+
+##### Server rules for scheduled colors
+
+Servers SHOULD NOT schedule an update more than 20 seconds ahead.
```
kahrendt: SHOULD NOT send a scheduled update more than 20 seconds before its timestamp

roles/metadata/v1.md
```diff
@@ -19,9 +19,21 @@ The `metadata` object in [`server/state`](../../messaging.md#server--client-serv
...
 
+#### Scheduled metadata updates
+
+A `metadata` object whose `timestamp` is in the future is a scheduled update: state that takes effect at that time (for example, the next track's metadata timed to the audible track change).
+
+Clients keep a **current state** plus at most one **pending update**. The current state is the running merge of applied updates and is what the client displays. A message whose `timestamp`, translated to the local clock via the [time filter](../../messaging.md#clock-synchronization) (current best estimate, no waiting for convergence), is still in the future becomes the pending update, replacing any held one, and is applied to the current state when that moment is reached. A message whose translated timestamp is in the past or present is applied immediately and discards any held pending update. Clients MAY show the pending update early (e.g. a coming-up display for the next track).
+
+##### Server rules for scheduled metadata
+
+Servers SHOULD NOT schedule an update more than 20 seconds ahead.
```
kahrendt: HOULD NOT send a scheduled update more than 20 seconds before its timestamp

roles/metadata/v1.md
```diff
@@ -19,9 +19,21 @@ The `metadata` object in [`server/state`](../../messaging.md#server--client-serv
...
+
 #### Calculating current track position
 
-Clients can calculate the current track position at any time using the `timestamp` and `progress` values from the last metadata message that included the `progress` object:
+Clients can calculate the current track position at any time using the `timestamp` and `progress` values from the most recent applied metadata that included the `progress` object (see [Scheduled metadata updates](#scheduled-metadata-updates)):
 
 ```python
 calculated_progress = metadata.progress.track_progress + (current_time - metadata.timestamp) * metadata.progress.playback_speed / 1000000
```
kahrendt, line 39:
```
calculated_progress = anchor.progress.track_progress + (current_time - anchor.timestamp) * anchor.progress.playback_speed / 1000000
```
Where `anchor` is the most recent applied metadata message that included `progress`.

maximmaxim345:
Marked as draft until #175 is merged.

roles/metadata/v1.md
maximmaxim345: #175 removed the "last message that included `progress`" concept.

kahrendt (approved):
LGTM!

PR #136, opened 2026-08-14
Allow local device controls to clear mute on volume change

maximmaxim345:
Requiring that any volume change preserve `muted` was confusing for a device whose only volume input is hardware buttons: pressing volume-up while muted would have to raise the volume and leave it silent. The note now covers only `server/command` and group volume commands, leaving local controls up to the device.

Not a breaking change since this PR just removes requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1250,7 +1250,7 @@ This section describes messages specific to clients with the `player` role, whic
 
 **Note:** To avoid audible clicks, clients SHOULD apply volume changes over a short ramp.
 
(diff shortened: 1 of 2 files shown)


kahrendt (approved):

mijofa:
Note the part of the spec you're changing here was already specifically referring to "via server command". So I believe this change is unnecessary for the use-case you've described of local device controls as it's not talking about them anyway.

EDIT: Wait, no, I might've misread the parenthised part...

PR #137, opened 2026-08-18
refactor: rename pairing PIN terminology to Pairing Code

chrisuthe:
Renames the pairing "PIN" concept to **Pairing Code** throughout the spec: prose, section headings and their anchors, and the wire protocol.

## Scope

76 occurrences across 4 source files, plus the regenerated `README.md`.

| File | Lines changed |
|---|---|
| `pairing.md` | 122 |
| `management.md` | 22 |
| `messaging.md` | 6 |
| `template.md` | 2 |
| `README.md` (generated) | 152 |

`reset-pinhole` in the Pairing Window section is deliberately untouched — that's a physical pin.

## ⚠️ Breaking: wire identifiers

| Before | After |
|---|---|
| `dynamic_pin` | `dynamic_pairing_code` |
| `static_pin` | `static_pairing_code` |
| `pin_length` | `pairing_code_length` |
| `min_pin_length` | `min_pairing_code_length` |
| `pin_mismatch` | `pairing_code_mismatch` |
| `pin_length_unacceptable` | `pairing_code_length_unacceptable` |
| `static_pin.pin` | `static_pairing_code.code` |

## ⚠️⚠️ Breaking, and it fails silently: the derivation label

The CPace pairing-code derivation label changes:

```
"sendspin-pin-derive-v1"  ->  "sendspin-pairing-code-derive-v1"
```

**The version suffix stays at `v1` while the derived value changes.** A peer updated on only one side derives a different code and fails key confirmation, surfacing as `pairing_code_mismatch` — indistinguishable from an operator typo.

This was a deliberate choice to keep the label consistent with the new terminology rather than leave a `pin` reference in the crypto. Calling it out prominently because a `v1` label that produces `v1`-incompatible output is the worst case for anyone debugging a pairing failure. Reviewers who would rather bump to `v2` (or leave the label untouched entirely) should say so — it is a one-line change either way.

The commit and PSK-wrap labels (`sendspin-pair-commit-v1`, `sendspin-pair-psk-wrap-v1`) are unchanged; they never contained "pin".

## ⚠️ Breaking: anchors

| Before | After |
|---|---|
| `#dynamic-pin-pairing-flow` | `#dynamic-pairing-code-flow` |
| `#static-pin-pairing-flow` | `#static-pairing-code-flow` |

External deep links to those two sections will break. All 14 in-repo references were updated.

## Prose conventions applied

- **Method names** are title-cased as proper nouns — "Dynamic Pairing Code", "Static Pairing Code" — matching the existing treatment of "Pairing PSK".
- **Flow references** use the full title-cased name including "Flow" ("the Dynamic Pairing Code Flow"), following the established `Pairing PSK Flow` precedent in running prose. Lowercase "flow" is kept only for pronouns ("this flow").
- **"code-based"** is used where the old text used "PIN" as a category label for the two PAKE methods as against Pairing PSK — matching the wording in the Definitions entry.
- **Derivation pseudocode** variables are lowercased to `code_int` / `code` / `derived_code` / `code_typed`, matching the surrounding snake_case field names.

## Verification

- `python3 tools/build-readme.py --check` passes. This is the meaningful check: the build fails on any link pointing at an anchor with no matching heading, so it proves every cross-reference to the two renamed headings resolves.
- Case-sensitive sweep for `PIN|\bpin\b|pin_|_pin\b|-pin-|\bPin\b` across all `.md` and `.py` files returns zero matches.


arturpragacz (approved):

maximmaxim345 (approved):
LGTM, thanks @chrisuthe !

PR #138, opened 2026-08-18
Replace `trust_level` with boolean `paired`

### LINKED ISSUE #140: Delete `trust_level` from `client/hello` instead of renaming it — the server already derives it from the matched PSK
## Problem

The message `client/hello` has the field `trust_level`. Its values are `'user'` and `'none'`.

The server can calculate the same value. The server knows which PSK matched during the Noise handshake.

This table shows the full mapping:

| Matched PSK | Trust |
|---|---|
| Sentinel PSK | unpaired |
| Pairing PSK | unpaired, during a pairing procedure |
| Long-term PSK | paired |

Both record types give the paired result. The spec says that "every record carries `user` trust level".

The spec already uses this calculation. management.md says that "the client validates that the matched PSK is a Sendspin PSK (i.e. the server is paired)". The table in `server/activate` uses the matched PSK, not the field `trust_level`.

connection.md makes each `psk_id` unique across the three categories. The reason is important: one `psk_id` must not map to two trust levels.

A client cannot correctly send a different value. The Definitions section makes the trust level a result of the record. If a client removes its record, the handshake fails before `client/hello`.

Thus the field is a second source of truth. The spec does not tell the server what to do if the two sources disagree.

## Change

Remove the field. Change these three locations:

1. messaging.md, `client/hello`. Remove the field description.
2. messaging.md, `server/unpair`. Change the condition to "If the matched PSK is not a long-term PSK". management.md already uses this style.
3. connection.md, Re-handshake. Remove the text "(the client re-asserts `trust_level`)".

PR #138 already changes the trust-level concept to paired and unpaired in approximately 8 more locations.

## Effect

The change removes one diagnostic signal. If a server uses the wrong PSK category, the value `'none'` gives an early indication.

After the change, the server finds the error at the `server/activate` rejection. The spec defines this rejection fully.

## Relation to PR #138

PR #138 changes the field to a boolean `paired`. This issue does more. If the server can calculate the value, the wire must not carry it.

Do one of these:

- Change PR #138 to remove the field.
- Merge PR #138, then apply this change.

This change breaks compatibility. The effect is equal to PR #138.</body>
</invoke>

maximmaxim345:
In a previous draft we had 3 trust levels: `none`, `user`, and `owner`. With only 2 left, a boolean `paired` says the same thing, and it keeps "trust level" jargon out of server and client UIs.

Keeping the field means the client can still correct a server whose own view of the pairing state has gone stale, which the matched PSK alone does not always reveal.

> [!WARNING]
> This is a breaking change for servers and clients.

Alternative to #158, which deletes the field entirely on top of per-category `psk_id` derivation labels.

Closes https://github.com/Sendspin/spec/issues/140


maximmaxim345:
Marked this PR as draft for now since #140 looks like an even better solution.

maximmaxim345:
Superseded by #158

maximmaxim345:
Closing since #158 was merged.

PR #155, opened 2026-08-18
Improve dynamic pairing

arturpragacz:
Improve dynamic pairing:
- Fix the dynamic pairing code at 6 digits.
- Define pairing code presentation format.
- Add QR-code emission format for the dynamic pairing code. Before the only QR-code format was defined for static Pairing PSK Flow, which is generally not the right tool to use for dynamic display purposes.
- Wrap `nonce_B` properly.
- Drop `other` from `out_channels`.
- Require Pairing PSK Token input format.

> [!WARNING]
> All of the changes here are breaking.


pairing.md
```diff
@@ -143,32 +122,40 @@ sequenceDiagram
...
 
-**Pairing code emission.** When emitting the pairing code through a spoken channel, the client SHOULD use the best-matching language it supports, treating the activation's [`languages`](messaging.md#server--client-serveractivate) as the language priority list under [RFC 4647](https://www.rfc-editor.org/rfc/rfc4647#section-3.4) Lookup matching, and falling back to its own default when nothing matches. The hint is informational and never grounds for [`pair/abort`](#client--server-pairabort); display emission is unaffected.
+**QR-code emission.** In the `qr_code` format the client presents the pairing code as a version-1 [pairing token](#pairing-token) with the 24-byte code as its payload (39 body characters), rendered as a QR code on its display. The server applies the token [decoding](#pairing-token) rules to operator input; the first 24 payload bytes are the entered pairing code.
```
maximmaxim345, line 142: I understand that this can improve the UX (of course assuming server and client are both capable of QR codes, but thats negotiated) and is mostly free to include in the spec, except defining the QR-code format in the spec.

But on the other hand, we already have a pairing code format. What would this add apart from it popping up automatically on the device when asked to scan it?

The point I'm not sure about is if we SHOULD require clients to support this though, this needs a QR-code dependency after all.

@kahrendt what do you think of this dynamic QR code idea?

pairing.md
```diff
@@ -218,7 +205,37 @@ sequenceDiagram
...
+
+Tokens are drawn only from the QR code alphanumeric set (`0–9`, `A–Z`, `:`), so they render as compact QR codes and survive manual transcription. A QR code carries the token string verbatim, with no URI scheme or wrapper, so a scan and a copy/paste yield identical input.
+
+Decoding reverses the transform and MUST be lenient with operator-supplied input:
+
+1. Trim surrounding whitespace and upper-case it.
+2. If present, strip a leading `SP:`. The first character is the `version`; reject an unrecognized version. An interface that accepts several versions dispatches on this character; one expecting a specific version rejects the others.
+3. Transliterate every `9` back to `2`, re-pad with `=` to a multiple of 8 characters, and base32-decode into the payload.
+
+A decoder MUST reject malformed input, including a payload shorter than its version defines. Payload bytes beyond those the version defines are reserved for future extension: a decoder MUST ignore them.
```
maximmaxim345, line 238: > Payload bytes beyond those the version defines are reserved for future extension

Why?
Don't we have `version` for exactly this reason?

pairing.md
```diff
@@ -67,47 +67,26 @@ Two standing client obligations follow from this flow:
...
-
-Decoding reverses the transform and MUST be lenient with operator-supplied input:
-
-1. Trim surrounding whitespace and upper-case it.
-2. If present, strip a leading `SP:`. The first character is the `version`; reject an unrecognized version.
-3. Transliterate every `9` back to `2`, re-pad with `=` to a multiple of 8 characters, and base32-decode.
-4. Reject the token unless the payload is exactly 64 bytes, then split it into `client_key` and `pairing_psk`.
-
-A decoder MUST reject malformed input. Before pairing, the server MUST confirm the decoded `client_key` matches the `client_id` presented on the connection.
+The operator enters the token into the server to begin the flow. The Pairing PSK MUST be exposed as the token, not the bare PSK. Before pairing, the server MUST confirm the decoded `client_key` matches the `client_id` presented on the connection.
```
maximmaxim345, line 79: Can you add this to the PR description?

pairing.md
```diff
@@ -143,32 +122,40 @@ sequenceDiagram
 **Binding values.** The Dynamic Pairing Code Flow introduces three values across two messages that bind the pairing code to the underlying Noise handshake:
 
 - `nonce_A` - 32 bytes drawn from a CSPRNG by the server, sent in [`server/pair-init`](#server--client-serverpair-init), base64url-encoded (43 chars).
-- `nonce_B` - 32 bytes drawn from a CSPRNG by the client, kept private until [`client/pair-confirm`](#client--server-clientpair-confirm) reveals it (base64url-encoded, 43 chars).
+- `nonce_B` - 32 bytes drawn from a CSPRNG by the client, kept private until [`client/pair-confirm`](#client--server-clientpair-confirm) reveals it as `wrapped_nonce_B`, [sealed under the CPace output](#wrapping).
```
maximmaxim345, line 125: What does wrapping it add?

pairing.md
```diff
@@ -67,47 +67,26 @@ Two standing client obligations follow from this flow:
...
 
-#### Pairing Token
-
-A server needs both the [Sendspin Pairing PSK](README.md#definitions) and the client's static public key to select and verify the client's Noise identity. The two are distributed together as a **pairing token**: a single case-insensitive ASCII string the operator transfers out of band (copy/paste, QR scan) into the server to begin the [Pairing PSK Flow](#pairing-psk-flow). A client offering `pairing_psk` SHOULD surface the token rather than the bare PSK.
-
-A token is a fixed `SP:` prefix, a version, and a base32-encoded body:
+**Pairing Token.** A server needs both the [Sendspin Pairing PSK](README.md#definitions) and the client's static public key to select and verify the client's Noise identity. The two are distributed together inside of a version-0 [pairing token](#pairing-token):
```
```suggestion
**Pairing Token.** A server needs both the [Sendspin Pairing PSK](README.md#definitions) and the client's static public key to select and verify the client's Noise identity. The two are distributed together in a version-0 [pairing token](#pairing-token):
```

pairing.md
```diff
@@ -13,11 +13,11 @@ This specification defines three pairing methods. Servers must implement all thr
...
 
-Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned.
+Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned. Clients whose display can render a QR code SHOULD offer also the `qr_code` [emission format](#dynamic-pairing-code-flow).
```
```suggestion
Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
```

Depends on how this discussion goes though: https://github.com/Sendspin/spec/pull/155/changes#r3805646510

maximmaxim345 (commented):
Left a few questions. Also, can you add a section to this PRs description roughly describing the breaking changes? Or at least add a:
> [!WARNING]
> so its clearer that this contains breaking changes?

pairing.md
arturpragacz: That way we don't have to define a new version for every single backwards-compatible extension, and most that I can envision in the future should be backwards-compatible.

pairing.md
arturpragacz: This was already there before, just with `SHOULD` instead of `MUST` (assuming you're referring to the second sentence here).

Added it to the PR description.

pairing.md
arturpragacz: It makes the security design consistent and ensures that all the CPace security properties apply, I missed that one before in #117.

pairing.md
arturpragacz: It offers both a better user experience and a better security design (dynamic secret vs static one, cleaner QR-code data).

`SHOULD` is not `MUST`, so if the QR dependency is too heavy, then I think that's a valid reason for the implementation to drop it. We can add some language in that direction though, if you think that's important.

maximmaxim345: Makes sense, allowing for better UX is a good enough reason alone to have it in the Spec.

> We can add some language in that direction though, if you think that's important.

It's fine as is, the `SHOULD` is good enough.

maximmaxim345 (approved):
Don't have any more questions/nits.
Thanks @arturpragacz !

PR #158, opened 2026-08-20
Delete `trust_level` from `client/hello`

### LINKED ISSUE #140: Delete `trust_level` from `client/hello` instead of renaming it — the server already derives it from the matched PSK

maximmaxim345:
The server already knows which PSK matched during the Noise handshake, so it can derive whether the session is paired instead of taking the client's word for it. Nothing consumed the field: every enforcement rule already keys on the matched PSK.

One gap made the field load-bearing until now: a management server could remove a record and then set its PSK as the client's pairing PSK, and the original server's next handshake would complete with the two sides disagreeing about the matched category - `trust_level` was the only signal exposing that. This PR therefore adds `psk_category: 'lt' | 'pr' | 'sn'` to the Noise message 1 payload: the server declares which category it is using the referenced PSK as, and the client matches `psk_id` only among candidates of that category. A recategorized credential is now an ordinary lookup miss that lands in the Sentinel fallback, which already tells the server to re-pair, so paired state is fully derivable on both sides and the field can go. A miss during a re-handshake instead fails the handshake as before; the server's reconnect then takes the initial-handshake fallback path.

Design notes: a category mismatch is deliberately a miss rather than a hard failure - the fallback turns it into an authenticated credential-mismatch signal with the re-pair path attached, where a hard failure would just produce a mute reconnect loop. The three codes share one length, keeping the encrypted payload's length independent of the category. The `psk_id` derivation, the published Sentinel constant, and the `already_exists` guards are all unchanged; the cross-category uniqueness rule stays, its rationale restated as what it actually protects (the pairing token distributes the pairing PSK in cleartext, so the same bytes must not also back a record).

This PR also rewords the "trust level" term in the spec. This was a left over from when we had 3 (the additional owner level, whose management rights were merged into the user level - now called just paired).

`unpaired_access.enabled` remains because it is genuine client policy the server cannot derive.

## Breaking change

Existing pairing records and `psk_id`s are untouched, so deployed pairings keep working. Breaking only when a client is updated before its server: an updated client requires `psk_category` in Noise message 1 and no longer sends `trust_level`, both of which an old server trips on. The reverse order is safe: receivers ignore unrecognized payload fields, so an updated server interoperates with old clients, which then simply don't enforce the category check.

---

Closes #140

Alternative to #138, which keeps the field as a boolean `paired` instead.


arturpragacz (changes requested):
After thinking about it some more I'm not sure it's a good idea to do this actually.

The problem is that there are some very weird edge cases, which are not going to happen in normal use, but someone could do it theoretically and so we have to handle them somehow. Consider for example:
- two servers, both already paired to the client
- the second one removes the pairing record for the first one
- it then sets the pairing PSK to the same PSK as the first server had
- now when the first server connects it thinks it is paired, but it actually isn't

So I think we should either go with the other solution (using `paired` field) or alternatively just leave it as is to make it easier to extend in the future if there is a need.

maximmaxim345:
Hmm, yea that's tricky to solve without any signal from the client

Maybe there's another approach if we change the `psk_id` derivation though: All three categories share one label right now, which is why a record and the pairing PSK can collide at all. Give each category its own label and the stale server just gets a lookup miss, which lands in the Sentinel fallback and already tells it to re-pair. I pushed a commit on this branch that tries this approach - it also means the `already_exists` guards compare PSK bytes instead of `psk_id`s, and the published Sentinel `psk_id` constant changes. The tradeoff: today a server could in principle detect the mismatch and silently re-pair itself in-session, while with the labels the mismatch is explicit and re-pairing goes through the normal flows - which after a deliberate `remove-record` seems like the right default.

The rename to `paired` is simpler though, so I reopened #138 as the alternative. In theory we can still expand later if we do want more "trust levels", just with an added `allowed_activities` field or similar instead of a third enum value. This is better backwards compatible as well: unknown fields have to be ignored, unknown enum values are undefined behavior or invalid I think.

@arturpragacz Let me know which you prefer - the `psk_id` labels here, or the boolean `paired` in #138.

arturpragacz:
That's overkill and makes things more complicated. We don't need domain separation here.

But I think I agree failing early is not a bad idea.

So instead revert the psk_id changes and just add:
```
psk_category: 'lt' | 'pr' | 'sn'
```
to Noise 1 message (keep it exactly as above for same length).

maximmaxim345:
Done, added the `psk_category` field instead in this PR.

Also, I made a category mismatch a lookup miss rather than a hard handshake failure, so it lands in the Sentinel fallback and the server gets the credential-mismatch signal and re-pair prompt instead of a silent reconnect loop.

messaging.md
```diff
@@ -404,7 +404,7 @@ Client behavior:
 
 - Remove the matched pairing record, send [`client/goodbye`](#client--server-clientgoodbye) reason `'unpaired'`, and close the connection.
 - If the matched record is a **shared-PSK record** (not bound to a `server_id`; may back other servers - see [Records](management.md#records)), the client MUST NOT remove it. It still sends `client/goodbye` reason `'unpaired'` and closes. Wholesale removal of a shared record requires [`management/remove-record`](management.md#server--client-managementremove-record).
-- If the connection's `trust_level` is `'none'` (e.g., an in-flight pairing handshake), ignore the message and continue unchanged.
+- If the session is [unpaired](README.md#definitions), there is no record to remove (e.g., an in-flight pairing handshake keyed by the pairing PSK), so ignore the message and continue unchanged.
```
arturpragacz:
```suggestion
- If the session is [unpaired](README.md#definitions), there is no record to remove, so ignore the message and continue unchanged.
```

README.md
```diff
@@ -694,7 +694,7 @@ Client behavior:
 
 - Remove the matched pairing record, send [`client/goodbye`](#client--server-clientgoodbye) reason `'unpaired'`, and close the connection.
 - If the matched record is a **shared-PSK record** (not bound to a `server_id`; may back other servers - see [Records](#records)), the client MUST NOT remove it. It still sends `client/goodbye` reason `'unpaired'` and closes. Wholesale removal of a shared record requires [`management/remove-record`](#server--client-managementremove-record).
-- If the connection's `trust_level` is `'none'` (e.g., an in-flight pairing handshake), ignore the message and continue unchanged.
+- If the session is [unpaired](#definitions), there is no record to remove (e.g., an in-flight pairing handshake keyed by the pairing PSK), so ignore the message and continue unchanged.
```
arturpragacz:
```suggestion
- If the session is [unpaired](#definitions), there is no record to remove, so ignore the message and continue unchanged.
```

arturpragacz (approved):

PR #160, opened 2026-08-20
refactor: make the binary message ID table normative, drop bit-layout prose

### LINKED ISSUE #144: messaging.md: drop the self-contradictory bit-layout prose for binary message IDs; make the assignment table normative
## Problem

The section "Binary Message ID Structure" gives a bit layout. The next sentence disagrees with it:

> Binary message IDs typically use **bits 7-2** for role type and **bits 1-0** for message slot, allocating 4 IDs per role. Roles with expanded allocations use **bits 2-0** for message slot (8 IDs).

The reader cannot decode an ID with this layout. The width of the slot field is 2 bits or 3 bits. The width depends on the role, and only the table below shows the role.

The layout also disagrees with the role numbers in the same section. The visualizer role uses IDs 16-23. Thus it fills role number 4 and role number 5 with the bits 7-2 rule that the label "Roles 6-47" uses.

The list "Message slots" repeats the table.

No part of the spec calculates a role from an ID with a bit mask. Each role gives its own IDs. An application-specific role gives its IDs in its own definition. The fragmentation section defines the bit 0 flag separately. No section refers to the role numbers.

## Change

Replace the first sentences and the list "Message slots" with this text:

````markdown
### Binary Message ID Structure

The first byte of every binary message is its message ID. IDs are assigned from the table below; each role's binary message definitions name the exact IDs it uses.

| IDs | Assignment |
|---|---|
| 0 | JSON message body (UTF-8) |
| 1 | Reserved for future use |
| 2-3 | [Fragmentation](#fragmentation) |
| 4-7 | Player role |
| 8-11 | Artwork role |
| 12-15 | Source role |
| 16-23 | Visualizer role |
| 24-191 | Reserved for future roles |
| 192-255 | Available for use by [application-specific roles](README.md#application-specific-roles) |

Future roles will be allocated aligned blocks of 4 or 8 IDs from the reserved 24-191 range.

**Note:** Role versions share the same binary message IDs (e.g., `player@v1` and `player@v2` both use IDs 4-7).
````

Make one related change. The file `roles/artwork/v1.md` says "Type `8`: Channel 0 (Artwork role, slot 0)". The term "slot" is then unnecessary. Keep the channel number only.

This change is editorial. All IDs stay the same. No source file refers to this anchor.</body>
</invoke>

### PR DESCRIPTION
**Pull request by chrisuthe**
Fixes #144.

Applies the replacement text from the issue verbatim: the self-contradictory bit-layout prose (bits 7-2 / bits 1-0, contradicted by the per-role slot widths and by the visualizer's 8-ID allocation) and the redundant "Message slots" list are replaced by the normative assignment table. Nothing in the spec decodes a role from an ID by bit mask; each role's binary message definitions name their exact IDs, and the table is now the single source.

The related `roles/artwork/v1.md` change is applied to all four channel lines, generalizing the issue's `Type 8` example:

```
- Type `8`: Channel 0 (Artwork role, slot 0)   ->   - Type `8`: Channel 0
```

(and likewise types 9-11). The two remaining uses of "slot" in the spec (`management.md` storage units, `pairing.md` digit-entry UI) are unrelated senses and untouched.

## Verification

- All IDs unchanged — the diff is prose/format only.
- Confirmed no source file links to `#binary-message-id-structure`, as the issue states.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,26 +335,21 @@ WebSocket binary messages are used to send JSON payloads, audio chunks, media ar
 
 ### Binary Message ID Structure
 
-Binary message IDs typically use **bits 7-2** for role type and **bits 1-0** for message slot, allocating 4 IDs per role. Roles with expanded allocations use **bits 2-0** for message slot (8 IDs).
+The first byte of every binary message is its message ID. IDs are assigned from the table below; each role's binary message definitions name the exact IDs it uses.
 
-**Role assignments:**
-- `00000000` (0): JSON message body (UTF-8)
-- `00000001` (1): Reserved for future use
-- `0000001x` (2-3): Used for [Fragmentation](#fragmentation)
-- `000001xx` (4-7): Player role
-- `000010xx` (8-11): Artwork role
-- `000011xx` (12-15): Source role
-- `00010xxx` (16-23): Visualizer role
-- Roles 6-47 (IDs 24-191): Reserved for future roles
-- Roles 48-63 (IDs 192-255): Available for use by [application-specific roles](#application-specific-roles)
+| IDs | Assignment |
+|---|---|
+| 0 | JSON message body (UTF-8) |
+| 1 | Reserved for future use |
+| 2-3 | [Fragmentation](#fragmentation) |
+| 4-7 | Player role |
+| 8-11 | Artwork role |
+| 12-15 | Source role |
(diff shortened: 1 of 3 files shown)


maximmaxim345 (approved):
Thanks @chrisuthe !

Issue #100, opened 2026-06-11
Tighter Key Word usage


jaredtrent:
Request for an editorial pass according to BCP 14 / RFC 2119 / RFC 8174.

https://www.rfc-editor.org/info/bcp14/

Focusing on more consistent and unambiguous language for requirements within the written spec -- words like MUST, MUST NOT, SHOULD, SHOULD NOT, and so on.

PR #161, opened 2026-08-20
refactor: move normative requirements out of Note blocks into body text

### LINKED ISSUE #147: Make `**Note:**` non-normative: relocate the 14 Notes carrying BCP 14 keywords into body text
## Problem

The spec defines the BCP 14 keywords in the section "Normative Language". But many requirements are in `**Note:**` blocks.

A note looks like a comment. The reader can omit it. But 14 of the 30 notes contain MUST, MUST NOT, SHOULD, SHOULD NOT or MAY.

One example is in `roles/player/v1.md`: "a volume change MUST NOT clear the mute state". Thus the reader cannot safely omit a note.

## Change

Use `**Note:**` for comments only. Move each sentence with a BCP 14 keyword into the body text.

Move the text. Do not change the keyword to lower case. A lower-case keyword gives a different problem: the requirement becomes unclear.

## Notes to move

These are the line numbers on `main`:

- `connection.md:19` — SHOULD. The TXT `name` and `client/hello`.
- `connection.md:54` — SHOULD. The TXT `name` and `server/hello`.
- `messaging.md:239` — SHOULD. The minimal activity set.
- `messaging.md:241` — MAY. The server can omit a role.
- `messaging.md:328` — MAY. The client can send unchanged fields again.
- `messaging.md:436` — SHOULD. The reason `shutdown`.
- `messaging.md:440` — SHOULD NOT. The reconnect backoff, in the list in the note.
- `roles/player/v1.md:4` — SHOULD. The volume-to-amplitude calculation.
- `roles/player/v1.md:6` — SHOULD. The volume ramp.
- `roles/player/v1.md:8` — MUST NOT. The volume change and the mute state.
- `roles/player/v1.md:25` — SHOULD, MAY. The format selection.
- `roles/source/v1.md:20` — MUST. The server supports all audio codecs.
- `roles/source/v1.md:22` — SHOULD. The server accepts the source format.
- `roles/source/v1.md:98` — SHOULD, SHOULD NOT. The source timestamps.

Some notes give advice at SHOULD level. One target section is already a large paragraph. Thus write the new text again. Do not only move it.

## Related problems

Some notes give a requirement with a lower-case keyword. The files `roles/artwork/v1.md:21` and `roles/player/v1.md:23` say "Servers must support …". The file `roles/source/v1.md:20` gives the same requirement with MUST. Thus the spec has three copies and two styles.

There are more examples at `messaging.md:381`, `messaging.md:440`, `roles/player/v1.md:27` and `roles/player/v1.md:86`.

The note at `messaging.md:243` has the opposite problem. It gives a requirement for the server, but it uses no keyword. To move this note is not sufficient.

Issue #100 asks for a general keyword pass. This issue gives a smaller subset with a mechanical test: a `**Note:**` block must not contain an upper-case BCP 14 keyword.

Refs #100.</body>
</invoke>

chrisuthe:
Fixes #147. Refs #100.

Relocates all 14 `**Note:**` blocks carrying BCP 14 keywords into body text, per the issue's list (located by content — line numbers had drifted with #137/#155). Keywords stay uppercase; wording is untouched except for the two integrations below. The issue's mechanical test now passes: **no `**Note:**` block contains an uppercase BCP 14 keyword** (16 comment-only notes remain).

## Straight promotions (11)

The `**Note:** ` marker is dropped and the paragraph stands as body text in place — each already sat adjacent to the content it governs:

- `connection.md` ×2 — TXT `name` / hello-name agreement (client and server sections)
- `messaging.md` ×4 — minimal activity set; MAY-omit-role; `server/state` asymmetry (clients MAY resend unchanged fields); disconnect-without-goodbye handling incl. the reconnect-backoff SHOULD NOT
- `roles/player/v1.md` ×2 — volume/mute independence (MUST NOT clear mute); initial format selection (SHOULD/MAY)
- `roles/source/v1.md` ×3 — servers MUST support all codecs; SHOULD accept the announced source format; source-timestamp handling (SHOULD/SHOULD NOT)

## Integrations (3 notes → 2 places)

Per the issue's "write the new text again — do not only move it":

- **`shutdown` note** folds into the `shutdown` reason bullet in `client/goodbye`, matching the style of the other reason bullets (which already carry normative sentences). "clients SHOULD send `shutdown`" becomes "clients SHOULD send this reason" now that it sits inside the bullet defining it.
- **The two adjacent volume notes** in `roles/player/v1.md` (perceived-loudness curve + click ramp) join into one body paragraph; both sentences govern the same behavior.

## Left alone, deliberately

Notes stating requirements with lowercase keywords (`roles/player/v1.md` "Servers must support…", `roles/artwork/v1.md`, and the keyword-free server requirement in the role-removal note) are outside this issue's mechanical test and belong to the general keyword pass in #100, as the issue states.

## Verification

- Sweep of every note block **including attached lists** (the backoff SHOULD NOT hides in a list continuation, not the note's first line): 14 found before, 0 after.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Clients announce their presence via mDNS using:
 
 The server discovers available clients through mDNS and connects to each client via WebSocket using the advertised address and path.
 
-**Note:** The TXT `name` SHOULD match the `name` the client sends in [`client/hello`](#client--server-clienthello). It is only a discovery-time hint; if the two differ, the `client/hello` value is authoritative.
+The TXT `name` SHOULD match the `name` the client sends in [`client/hello`](#client--server-clienthello). It is only a discovery-time hint; if the two differ, the `client/hello` value is authoritative.
 
 **Note:** Do not manually connect to servers if you are advertising `_sendspin._tcp`.
 
@@ -189,7 +189,7 @@ If clients prefer to initiate the connection instead of waiting for the server t
 
 Clients discover the server through mDNS and initiate a WebSocket connection using the advertised address and path.
 
-**Note:** The TXT `name` SHOULD match the `name` the server sends in [`server/hello`](#server--client-serverhello). It is only a discovery-time hint; if the two differ, the `server/hello` value is authoritative.
+The TXT `name` SHOULD match the `name` the server sends in [`server/hello`](#server--client-serverhello). It is only a discovery-time hint; if the two differ, the `server/hello` value is authoritative.
 
 **Note:** Do not advertise `_sendspin._tcp` if the client plans to initiate the connection.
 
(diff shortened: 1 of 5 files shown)


maximmaxim345 (approved):
Looks good! Noticed that some other notes that don't use the RFC style wording can also benefit from such a pass. I'll open a separate follow up PR for that.
Thanks @chrisuthe !

Issue #141, opened 2026-08-18
Use one canonical name per PSK category (long-term PSK / pairing PSK / sentinel PSK)


balloob:
## Problem

The three PSKs are difficult to identify. This is the most complex part of the spec. Two of the three PSKs have more than one name.

connection.md already gives the category names: "(long-term, Pairing, Sentinel)". The prose does not use them.

## Change

Use one name for each category. The names agree with the wire identifiers.

| Current name | Locations | New name |
|---|---|---|
| Sendspin PSK | 7 | long-term PSK |
| long-term Sendspin PSK | 6 | long-term PSK |
| long-term PSK | 5 | no change, becomes the standard name |
| per-server PSK | 2 | long-term PSK |
| per-pair Sendspin PSK | 1 | long-term PSK |
| Sendspin Pairing PSK | 6 | pairing PSK |
| Pairing PSK | 24 | pairing PSK |
| Sentinel PSK | 18 | no change |

The change affects approximately 22 locations in template.md, connection.md, messaging.md, pairing.md and management.md. The role files do not refer to a PSK.

The change does not affect the wire. The names agree with `long_term_psk`, `pairing_psk` and the constant `sendspin-sentinel-psk-v1`.

## Alternatives that we do not recommend

Do not remove the suffix "PSK", for example "pairing key". These keys are the Noise PSK input to `KKpsk2`. The `psk_id` procedures use the term "PSK". The name "pairing key" also points to the wrong PSK.

Do not change "Sentinel" to "unpaired PSK". The term is correct, and all 18 locations use it. It also agrees with the published label constant.

## Correction to make at the same time

Do not use "per-pair" as the standard name. With a shared-PSK record, one long-term PSK can authenticate more than one server.

The definition "shared between a (client, server) pair" has the same error.</body>
</invoke>

arturpragacz:
Makes sense.

PR #162, opened 2026-08-20
refactor: drop the Sendspin prefix from defined terms

### LINKED ISSUE #148: Drop the "Sendspin" prefix from defined terms where it carries no information
## Problem

Almost all items in this spec are Sendspin items. Thus the prefix "Sendspin" on a defined term adds length, but it adds no data.

The body text already agrees. The bare words "client", "server", "group" and "stream" occur more than 1000 times. All five references to the trust level omit the prefix. The terms "long-term PSK" and "Pairing PSK" also occur without the prefix.

## Change

Remove the prefix from each term below. Do this term by term, not with one general rule.

| Term in the Definitions | Change | Reason |
|---|---|---|
| Sendspin Server | **Server** | The body text uses "server" approximately 484 times. Keep the prefix in the discovery text in connection.md. |
| Sendspin Client | **Client** | The body text uses "client" approximately 468 times. Keep the prefix in two locations in connection.md, see below. |
| Sendspin Group | **Group** | The spec has no other type of group. |
| Sendspin Stream | **Stream** | The prefix occurs only in the definition. The body text uses "stream" approximately 85 times. |
| Sendspin Identity | **Identity** | No other section refers to this term. |
| Sendspin PSK | **long-term PSK** | Do not use the bare term "PSK", see below. |
| Sendspin Pairing PSK | **Pairing PSK** | pairing.md and management.md already use the bare term. |
| Sendspin Pairing Code | **Pairing Code** | pairing.md already uses the bare term. Only the definition links use the prefix. |
| Sendspin Trust Level | **Trust Level** | All five references already use the bare term. |

## Two exceptions

Do not use the bare term "PSK" for the long-term PSK. The spec has three PSKs, and the `server/activate` table shows all three together. The spec also uses "PSK" and `psk_id` in a general sense. Thus change this term to "long-term PSK". The spec already uses this name in five locations.

Keep the prefix in connection.md at two locations. With a server-initiated connection, the Sendspin client is the WebSocket listener. Thus "Sendspin client" and "WebSocket client" are different, and the prefix is necessary.

Also keep the prefix in the adjective form. The section External Source Handling uses "Sendspin playback" and "non-Sendspin activity". Here the difference is the function of the sentence. The file `roles/source/v1.md` does the same.

## Notes

This change is editorial. It does not affect the wire. It does not change a message type, a field name, or the service types `_sendspin._tcp` and `_sendspin-server._tcp`.

All definition links point to the section anchor `README.md#definitions`. Thus no link breaks.

README.md is a generated file. Generate it again after the change.

Issue #141 also changes "Sendspin PSK" to "long-term PSK". Complete one issue, then remove this row from the other issue.</body>
</invoke>

chrisuthe:
Fixes #148.

Drops the "Sendspin" prefix term by term per the issue's table: **Server**, **Client**, **Group**, **Stream**, **Identity**, **Pairing Code**, and **Trust Level** in the Definitions block, plus the prefixed body references ("Sendspin group" ×3, "Every Sendspin client and server", "the Sendspin client implements", "a different Sendspin server", and the discovery-chapter opener "Sendspin Servers must support both methods").

## Kept, per the issue's exceptions

- `connection.md:13` and the connection note (`:62`) — "Sendspin client" where the Sendspin client is the WebSocket listener / always the data consumer, so the prefix disambiguates against the WebSocket role.
- `connection.md:48` — server discovery text.
- The adjective forms: "Sendspin playback" ×3 and "non-Sendspin activity" ×2 in External Source Handling, and the system-name use in `roles/source/v1.md`.

## Coordination with #159

The issue's two PSK rows (Sendspin PSK → long-term PSK, Sendspin Pairing PSK → pairing PSK) are **not** in this PR — they are implemented by #159 (issue #141), as this issue anticipates ("Complete one issue, then remove this row from the other issue"). The two PRs touch adjacent lines of the Definitions block and both regenerate `README.md`, so whichever merges second may need a trivial rebase + `python3 tools/build-readme.py`.

## Verification

- Remaining `Sendspin <term>` occurrences in source files are exactly the exception list above — audited by grep.
- No wire identifiers, message types, or the `_sendspin._tcp` / `_sendspin-server._tcp` service types touched.
- `python3 tools/build-readme.py --check` passes; definition links all target the unchanged `#definitions` anchor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,22 +91,22 @@ sequenceDiagram
 
 ## Definitions
 
-- **Sendspin Server** - orchestrates all devices, generates audio streams, manages players and clients, provides metadata
-- **Sendspin Client** - a client that can play audio, capture audio inputs, visualize audio, display metadata, display colors, or provide music controls. Has different possible roles (player, source, metadata, controller, artwork, visualizer, color). Every client has a unique identifier
+- **Server** - orchestrates all devices, generates audio streams, manages players and clients, provides metadata
+- **Client** - a client that can play audio, capture audio inputs, visualize audio, display metadata, display colors, or provide music controls. Has different possible roles (player, source, metadata, controller, artwork, visualizer, color). Every client has a unique identifier
   - **Player** - receives audio and plays it in sync. Has its own volume and mute state and preferred format settings
   - **Source** - captures audio from a local input and streams it to the server
-  - **Controller** - controls the Sendspin group this client is part of
+  - **Controller** - controls the group this client is part of
   - **Metadata** - displays text metadata (title, artist, album, etc.)
   - **Artwork** - displays artwork images. Has preferred format for images
   - **Visualizer** - visualizes music. Has preferred format for audio features
   - **Color** - receives colors derived from the current audio
-- **Sendspin Group** - a group of clients. Each client belongs to exactly one group, and every group has at least one client. Every group has a unique identifier. Each group has the following states: list of member clients, volume, mute, and playback state
-- **Sendspin Stream** - client-specific details on how the server is formatting and sending binary data. Each role's stream is managed separately. Each client receives its own independently encoded stream based on its capabilities and preferences. For players, the server sends audio chunks as far ahead as the client's buffer capacity allows. For artwork clients, the server sends album artwork and other visual images through the stream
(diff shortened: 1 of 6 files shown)


maximmaxim345 (approved):
Thanks @chrisuthe !

PR #163, opened 2026-08-20
refactor: rename client_stream/* to client-stream/*, document prefix rule

### LINKED ISSUE #150: Rename `client_stream/*` to `client/stream-*` and document the message-type naming convention
_Revised after the review by @maximmaxim345. The first version proposed `client/stream-start`, and that name was wrong. See the discussion below._

## Problem

The prefix `client_stream/` is the only message prefix with an underscore. Two messages use it, `client_stream/start` and `client_stream/end`.

The spec also does not give the rule for its message prefixes. Thus the reader makes a rule, and then finds the exceptions:

- `stream/request-format` goes from the client to the server. The other `stream/` messages go from the server to the client.
- `management/` requests go from the server to the client. Only `management/result` goes to the server.
- `pair/abort` and `noise/handshake` go in both directions.

## Change 1: rename `client_stream/*`

Change `client_stream/start` to `stream/client-start`. Change `client_stream/end` to `stream/client-end`.

The prefix `stream/` is a category, not a sender. @maximmaxim345 gives the rule: these messages carry the metadata for a binary channel, and they open, close or reconfigure that channel. The two `client_stream/*` messages have the same function.

Thus the new names keep the category. The prefix gives the function, and the name gives the direction. The messages `stream/start`, `stream/end` and `stream/clear` do not change.

All 15 occurrences are in `roles/source/v1.md`, and no other file links to them. Thus this is a change to one file.

## Change 2: give the rule in messaging.md

Add a short note in the Communication section, near the text about `type` and `payload`.

Give this rule: a prefix identifies a group of messages. The prefixes `client/` and `server/` give the sender. The prefix `stream/` gives the messages that control a binary channel. The prefixes `group/`, `management/`, `pair/` and `noise/` give a subject. A message that is not in the `client/` or `server/` group can go in either direction, and each message gives its own direction.

The rule for `stream/` is the important part. The spec does not give it today, and for this reason the first version of this issue proposed the wrong names.

## Not proposed

A full correction of the message names. It changes 16 to 22 of the 38 message types. It also cannot be fully consistent, because a message in both directions has no sender prefix.

@maximmaxim345 gives one alternative: rename `stream/start` to `server/stream-start`, and use `client/stream-start` for the source. This makes the sender the rule for all messages, but it removes the `stream/` category. This issue does not propose it.

## Compatibility

Change 1 breaks compatibility for a source client. No other message type changes.

chrisuthe:
Fixes #150.

_Revised after review: names are now `client-stream/*`, per the [final discussion on #150](https://github.com/Sendspin/spec/issues/150#issuecomment-5339878519). The issue body still shows the superseded `stream/client-*` form._

## Change 1: `client_stream/*` → `client-stream/*`

`client_stream/start` → `client-stream/start`, `client_stream/end` → `client-stream/end`. The underscore goes; the messages keep their own prefix rather than joining `stream/`, because server→client and client→server binary channels are independent and have separate lifetimes. All 18 occurrences are in `roles/source/v1.md`; the two section headings move with their anchors and in-file links are updated (build-verified). No other file links to these sections.

**⚠️ Breaking:** these two message types change on the wire for clients implementing the `source` role. No other message type changes.

## Change 2: the prefix rule

Added to the Communication section of `messaging.md`, directly after the `type`/`payload` paragraph:

> **Message type prefixes.** The prefix before the `/` in a message `type` identifies a group of messages. `client/` and `server/` name the sender. `stream/` groups the messages that control a binary channel from the server to the client, and `client-stream/` those that control a binary channel from the client to the server; the two kinds of channel are independent and have separate lifetimes. `group/`, `management/`, `pair/`, and `noise/` name a subject. A message outside the `client/` and `server/` groups can flow in either direction; each message's definition gives its direction.

The closing sentence is what places `stream/request-format` (client-sent, but controlling the server→client channel) correctly. Body text rather than a `**Note:**`, in keeping with #147.

## Coordination

#161 (open) touches `roles/source/v1.md` on one of the same lines (the promoted format-announcement note). Whichever merges second needs a trivial rebase + README rebuild — happy to do that on this side.

## Verification

- Zero `client_stream` or `stream/client-` occurrences remain in any source file.
- `python3 tools/build-readme.py --check` passes — the renamed headings' anchors and all links to them resolve.


maximmaxim345 (changes requested):
Can you adjust this PR to instead call it `client-stream/start` and `client-stream/end`? I think this got mixed up since the description of #163 still suggested `stream/client-start`.

See https://github.com/Sendspin/spec/issues/150#issuecomment-5339878519

maximmaxim345 (approved):
Thanks @chrisuthe !

Issue #142, opened 2026-08-18
player@v1: consolidate `supported_commands` into `client/state` and give the timing/buffering server rules their own section


balloob:
The file `roles/player/v1.md` has two structural problems. They are in the same file, but you can merge them separately.

## 1. Two messages contain `supported_commands`

`client/hello` contains `'volume'` and `'mute'` in the object `player@v1_support`. `client/state` contains `'set_static_delay'` in the object `player`.

Thus `server/command` must refer to two locations. It says that the command must be in `supported_commands` "from `player@v1_support` **or** from `client/state`".

The rules for `volume?` and `muted?` in `client/state` also refer to the `client/hello` anchor.

**Change:** move the full list into the object `player` in `client/state`.

This location is correct, because the client can change the list. The spec already permits a change to `supported_commands` when the audio output changes. But today the client can change only `'set_static_delay'`.

The change also adds a function. A client can then remove volume control during a session. @maximmaxim345 gives this example:

- The client plays through a built-in speaker, and it supports volume.
- The user connects an amplifier through line-out or S/PDIF. This output has no volume control.
- The client sends a new `supported_commands` list. The client does not disconnect.

No procedure needs the list before the first `client/state`. The server must not send binary data before this message. The player is not `available` before this message. The values for volume and mute are also in this message.

The controller calculates the group volume. This calculation needs the flags and the values, and `client/state` supplies the values.

Make these two related changes:

- Make `supported_commands` mandatory in the first `client/state`. The delta rules permit its omission later.
- Add one sentence about the controller. The server must calculate the controller `supported_commands` again when the support of a player changes.

This change breaks compatibility. The field moves to a different message.

## 2. The "Server behavior" text is below the incorrect title

The text is below the title for the `client/hello` support object. But it describes `required_lead_time_ms` and `min_buffer_ms`. These fields are in `client/state`.

Thus `client/state` refers back to the `client/hello` anchor for its own fields. The Playback Synchronization section does the same.

**Change:** move the text to a new section, for example "Server timing and buffering behavior". Move the note before it at the same time.

Keep the new section independent. Do not put it in the `client/state` section. The text also gives rules for `buffer_capacity`, and this field stays in `client/hello`.

Two links change. The links to `supported_formats` continue to operate, because the `client/hello` title does not change.

This change is editorial. It does not affect the wire.

maximmaxim345:
[...]

Issue #93, opened 2026-05-27
Static delay: specify what a client does with an out-of-range `static_delay_ms`


balloob:
_This issue had two parts. The spec already gives the persistence requirement. Only the range check stays. See the comment below._

## Problem

The spec gives the range of `static_delay_ms` two times. The `client/state` player object says "static delay in milliseconds (0-5000)". The `set_static_delay` parameter in `server/command` says "(0-5000)".

But the spec does not tell the client what to do with a different value. Thus the behavior is not defined. A defective server or a local configuration can send a value out of the range.

The SDKs are different. `sendspin-cpp` obeys the range 0-5000. `sendspin-js` limits the value to the range. `sendspin-cli` changes a negative value to a positive value, thus −100 becomes +100. This makes an audible error. The other SDKs do not examine the value.

## Change

Add this text to the paragraph "Static delay" in `roles/player/v1.md`:

> Clients MUST clamp a received `static_delay_ms` to the range 0–5000 and SHOULD log an out-of-range value as a protocol error.

The client limits the value. The client does not reject the message. Thus one incorrect command cannot stop the audio output. The log entry helps the operator to find the defect in the server.

The spec already says that "Negative values are not supported and should never be required for any compliant implementation". This sentence then has a defined result.

## Coordination

Issue #154 renames this field to `output_delay_ms`. If that issue lands first, use the new name in the text above.

---

**Source audit doc:** [`docs/static-delay.md`](https://github.com/Sendspin/conformance/blob/claude/stream-sync-correction-sdks-AWoNC/docs/static-delay.md)
**Full audit branch:** [`claude/stream-sync-correction-sdks-AWoNC`](https://github.com/Sendspin/conformance/tree/claude/stream-sync-correction-sdks-AWoNC/docs)
**Per-SDK digest:** [sdk-issues digest](https://gisthost.github.io/?ac84f17e71e7eccc48c9f08d5df0a7b4)

balloob:
Re-audited and re-scoped to item 2.

**Item 1 (persistence) — dropped.** The protocol-level requirement is already in `roles/player/v1.md`: "Clients must persist `static_delay_ms` locally across reboots and server reconnections", along with the per-output guidance ("persisting separate delays per output"). What this issue additionally asked for was a *recommended SDK persistence hook shape* — an SDK API concern rather than wire-protocol text, and one of several such asks across the May audit batch. Better tracked in the SDK repos or `Sendspin/conformance`.

[...]

PR #164, opened 2026-08-20
refactor: rename static_delay_ms to output_delay_ms

### LINKED ISSUE #154: Rename `static_delay_ms` to `output_delay_ms` (and `set_static_delay` to `set_output_delay`)
## Problem

The player role has `static_delay_ms` and the command `set_static_delay`. I still think the name is technically correct: it describes the static delay of the audio pipeline outside of what's controllable by the client (external speakers, amplifiers), as opposed to the delays the sync loop measures and adapts continuously.

But in practice it confuses users. For a user there is only one configurable delay, and it's the one called "static". They see a contradiction and ask why the delay is "static" when they can change it themselves.

## Change

Rename to "output delay". It describes the same thing (the latency of the output chain past the audio port) without the word that reads as "unchangeable", and it matches what users know from TVs and AVRs (audio delay, lip-sync delay).

Wire changes, all in `roles/player/v1.md`:

- `client/state` player object: `static_delay_ms` becomes `output_delay_ms`
- `supported_commands` value: `'set_static_delay'` becomes `'set_output_delay'`
- `server/command`: the `command` enum value and the payload field `static_delay_ms?` become `set_output_delay` and `output_delay_ms?`

Prose changes, same file:

- The bold "**Static delay:**" definition becomes "**Output delay:**"
- The references in the lead-time note, the `required_lead_time_ms` rule, the timestamp translation text, and the two Playback Synchronization bullets

Keep the existing distinction in the definition: this value covers delay beyond the audio port only, not processing delays (DAC latency, buffers) that the client compensates itself. One clarifying phrase there prevents the new name from being read as covering those.

No other source file refers to the field.

## Compatibility

This change breaks compatibility. It changes one `client/state` field, one command enum value, and one command payload field. There is no alias.

README.md is a generated file. Generate it again after the change.

## Coordination

- #142 moves `supported_commands`, which enumerates this command. Whichever change lands second uses the new name.
- #93 specifies out-of-range handling for this field. Same rule.

chrisuthe:
Fixes #154.

Renames the player role's static delay to **output delay** throughout `roles/player/v1.md` — field, command enum, `supported_commands` value, definition heading, and every prose reference (lead-time note, send-ahead formula, timestamp translation, both Playback Synchronization bullets). Verified no other source file refers to the field, as the issue states.

**⚠️ Breaking:** `static_delay_ms` → `output_delay_ms` (one `client/state` field, one `server/command` payload field) and `set_static_delay` → `set_output_delay` (command enum value + `supported_commands` entry). No alias. This affects every player client.

## The clarifying phrase

Per the issue ("One clarifying phrase there prevents the new name from being read as covering those"), the definition now reads:

> **Output delay:** The default is 0, meaning audio exits the device's audio port at the timestamp. `output_delay_ms` compensates for additional delay beyond the port (external speakers, amplifiers); it does not cover processing delays before the port (DAC latency, audio buffers), which the client compensates itself.

## One small fix beyond the issue's letter

The timestamp-translation text listed **amplifier delays** among the processing delays clients compensate themselves — contradicting the definition, which places amplifiers beyond the port under `output_delay_ms`. With the boundary now stated explicitly in the definition, the example list drops "amplifier delays":

> Clients should compensate for any known processing delays (e.g., DAC latency, audio buffer delays) by accounting for these delays when submitting audio to the hardware.

Happy to revert if amplifiers were intended on both sides.

## Coordination

Per the issue: #142 (moves `supported_commands`) and #93 (out-of-range handling for this field) should use the new name if they land after this.

## Verification

- Zero `static_delay` / "static delay" occurrences remain in any source file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1281,10 +1281,10 @@ The `player@v1_support` object in [`client/hello`](#client--server-clienthello) 
 
 **Note:** For the initial [`stream/start`](#server--client-streamstart) the server SHOULD select the highest-priority `supported_formats` entry it can produce for the current track. It MAY select a lower-priority entry when warranted, for example to match a track's native sample rate and avoid resampling, and MAY switch formats on a later track by sending a new `stream/start`.
 
-**Note:** [`required_lead_time_ms`](#client--server-clientstate-player-object) and [`min_buffer_ms`](#client--server-clientstate-player-object) are reported via [`client/state`](#client--server-clientstate-player-object). Players should report the lowest values that reliably prevent buffer underruns and start-of-stream truncation under expected conditions, to ensure the lowest possible latency for real-time applications. Both should factor in expected network delay/jitter (small on LAN/Wi-Fi, larger for remote or high-latency clients). Do not include `static_delay_ms` in these values; the server applies `static_delay_ms` separately when calculating send-ahead.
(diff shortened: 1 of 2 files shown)


maximmaxim345 (approved):
Thanks @chrisuthe !
The "One small fix beyond" also looks good.

Issue #90, opened 2026-05-27
Document PCM byte layout as its own subsection, de-duplicated across the player and source roles


balloob:
_This issue had two parts. Part 1 is withdrawn after the comment from @maximmaxim345. Only the PCM documentation stays._

## Problem

The spec gives the PCM byte layout in one sentence. The sentence is easy to miss.

The same sentence occurs in two files, `roles/player/v1.md` and `roles/source/v1.md`:

> **PCM Encoding Convention:** For the `pcm` codec, samples are encoded as little-endian signed integers (two's complement). 24-bit samples are packed as 3 bytes per sample.

The **Codec framing** text below it is also in the two files. It gives the rules for `pcm`, `flac` and `opus`, and a note about `codec_header`. Only the message name is different: `stream/start` or `client_stream/start`.

## Change

Move the two blocks into one new section. The section gives the PCM byte layout and the codec framing rules.

Then the player role and the source role refer to this section. Each role gives only its own format message.

Add an example for 16-bit and for 24-bit. Then a reader can examine an implementation against the example. Do not add a 32-bit example, because the spec gives only 16 and 24 as values for `bit_depth`.

## Reason

The file `sendspin-go` has no code for 24-bit samples. The current text is easy to miss, thus the error is difficult to find.

The change also removes a duplicate requirement. Today the team must keep two files in agreement.

## Part 1: withdrawn

Part 1 asked for a default codec order. The spec already says that `supported_formats` is in order of preference. Thus the client selects the order.

The user can also select PCM in the client settings. A SHOULD for Opus is then incorrect. A SHOULD that a correct implementation must ignore has no function.

If an SDK sends PCM first with no user setting, correct the default in that SDK.

---

**Source audit doc:** [`docs/codec-format-negotiation.md`](https://github.com/Sendspin/conformance/blob/claude/stream-sync-correction-sdks-AWoNC/docs/codec-format-negotiation.md)
**Full audit branch:** [`claude/stream-sync-correction-sdks-AWoNC`](https://github.com/Sendspin/conformance/tree/claude/stream-sync-correction-sdks-AWoNC/docs)
**Per-SDK digest:** [sdk-issues digest](https://gisthost.github.io/?ac84f17e71e7eccc48c9f08d5df0a7b4)</body>
</invoke>

maximmaxim345:
> ### 1. Specify a default codec ordering
> 
> > Clients SHOULD list `supported_formats` in descending preference,
[...]

PR #165, opened 2026-08-20
docs: record the editorial rules in CONTRIBUTING.md

### LINKED ISSUE #153: CONTRIBUTING.md: record the editorial rules for the split spec files
## Problem

The spec has editorial rules that the files obey, but CONTRIBUTING.md does not give them. Thus a contributor cannot see the rule, and proposes a change against it.

One rule is important: each role file must be self-contained. A person who implements the visualizer role and the metadata role must not read the PCM rules. For this reason the PCM byte layout and the codec framing rules occur two times, in `roles/player/v1.md` and in `roles/source/v1.md`.

This duplication looks like an error. Issue #90 proposed a shared section for these two blocks, and #90 was wrong for this reason. See the discussion in that issue.

## Change

Add a short section to CONTRIBUTING.md with the editorial rules. Give at least this rule:

> Each role file must be self-contained. A reader who implements one role must not read another role file. Thus text that applies to more than one role can occur more than one time. Do not combine such text into a shared section.

Add the other rules that the files obey today. Possible examples:

- README.md is a generated file. Change the source files only.
- Each heading must have a unique anchor, because the build script tests this.
- Where a term has more than one name, use the name in the Definitions section.

## Result

A contributor sees the rules before the work starts. A reviewer can also point to a rule, and does not repeat the explanation.

---

_This issue comes from the discussion in #90._

chrisuthe:
Fixes #153.

Adds an **Editorial rules** section to `CONTRIBUTING.md`, placed between the source-files section and the pre-commit hook. Six rules:

1. **Each role file is self-contained** — the issue's rule, with the PCM/codec-framing duplication across `roles/player/v1.md` and `roles/source/v1.md` named explicitly so the duplication stops looking like an error (the #90 trap).
2. **Edit the source files, never `README.md`** — the issue's first example.
3. **Every heading needs a unique anchor** — the issue's second example, stating both build failures (duplicate anchors, dangling links).
4. **Use one canonical name per term** — the issue's third example, extended with "prose names agree with the wire identifiers they describe" per the reasoning in #141/#148, plus the exception #148 kept for prefixes that disambiguate (`Sendspin client` vs the WebSocket client).
5. **`**Note:**` blocks are non-normative** — the mechanical rule from #147: uppercase BCP 14 requirements belong in body text.
6. **A client describes itself across `client/hello` and `client/state`**: @maximmaxim345's rule from the #153 comment, with the boundary set at the connection's lifetime rather than permanence.

Rules 4 and 5 now describe `main` as it stands. Rule 6 does not yet: `supported_commands` sits in both messages until #142 consolidates it.

No source spec file changes; `python3 tools/build-readme.py --check` still passes (CONTRIBUTING.md is not part of the README build).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ to intra-page anchors at build time.
 The source files follow these rules. Check a change against them before proposing it; reviewers point to them instead of re-explaining.
 
 - **Each role file is self-contained.** A reader who implements one role must not need to read another role file. Text that applies to more than one role therefore appears more than once - the PCM encoding convention and codec framing rules, for example, appear in both `roles/player/v1.md` and `roles/source/v1.md`. Do not combine such text into a shared section.
+- **A client describes itself across `client/hello` and `client/state`.** A field expected to change during a connection belongs in `client/state`; a field expected to stay constant for the connection, such as device identity or a fixed hardware limit, belongs in `client/hello`. `client/hello` is sent once per connection, so a field placed there can only be updated by reconnecting.
 - **Edit the source files, never `README.md`.** `README.md` is generated; the pre-commit hook regenerates it and blocks direct edits.
 - **Every heading needs a unique anchor.** The build fails on two headings that produce the same anchor, and on a link to an anchor with no matching heading.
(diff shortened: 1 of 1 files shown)


maximmaxim345 (approved):
Thanks @chrisuthe !

I also added another rule to decide if something belongs into `client/hello` or `client/state`.

PR #166, opened 2026-08-20
Move the remaining requirement-shaped notes into body text

maximmaxim345:
Follow-up to #161, which relocated the 14 `**Note:**` blocks carrying uppercase BCP 14 keywords (#147). Some notes that do not use the RFC style wording benefit from the same pass: they state a requirement with a lowercase keyword or with no keyword at all, so a reader still cannot safely skip a note. Refs #100.

Two moves, decided per note.

## Promoted to body text with an uppercase keyword

- The audio-codec and image-format support requirements in `roles/player/v1.md` and `roles/artwork/v1.md`. #161 promoted the `roles/source/v1.md` twin, so this class of requirement was body text in one role file and a lowercase note in two others. All three now match.
- The role-removal ordering rule in `messaging.md`, which had no keyword at all (now MUST). #147 notes that moving this one is not sufficient.
- Image scaling in `roles/artwork/v1.md` (now MUST). This is the only statement of that server behavior anywhere in the spec, so the requirement lived solely inside a note. The channel-numbering half of the note stays a note.
- The lead-time and buffer reporting guidance in `roles/player/v1.md` (SHOULD, and MUST NOT for the `static_delay_ms` exclusion). It sits directly above the Server behavior list that governs the same fields.

## Reworded so they read as comments

- The two `stream/request-format` notes, in `messaging.md` and `roles/player/v1.md`: "clients should use" becomes "clients can use". Purely advisory.

The two copies are kept because they sit at different scopes, the generic message and the player-specific object, matching how the role files already restate codec details independently.

## Left alone

`connection.md` has two bare-imperative notes ("Do not manually connect to servers if you are advertising `_sendspin._tcp`"). Both are one direction of the exclusivity MUST at the top of that section, so the rule is already stated normatively and the notes read as practical reminders.

Notes carrying a lowercase "may" in `messaging.md` (the `?` notation, and per-version support objects) define notation and naming rather than granting permission, with the requirements stated separately nearby.

## Verification

Every remaining `**Note:**` block in the source files is a comment with no requirement-shaped wording. `python3 tools/build-readme.py --check` passes.


README.md
```diff
@@ -1269,11 +1268,11 @@ The `player@v1_support` object in [`client/hello`](#client--server-clienthello)
...
+[`required_lead_time_ms`](#client--server-clientstate-player-object) and [`min_buffer_ms`](#client--server-clientstate-player-object) are reported via [`client/state`](#client--server-clientstate-player-object). Players SHOULD report the lowest values that reliably prevent buffer underruns and start-of-stream truncation under expected conditions, to ensure the lowest possible latency for real-time applications. Both SHOULD factor in expected network delay/jitter (small on LAN/Wi-Fi, larger for remote or high-latency clients). These values MUST NOT include `static_delay_ms`; the server applies `static_delay_ms` separately when calculating send-ahead.
```
kahrendt: Nit: "These values MUST NOT"
MUST NOT is the actionv, so the player should be the one following this, the values have no agency on their own

kahrendt (commented):
I think this needs a rebase still

README.md
maximmaxim345: Rebased and reworded that sentence in a5a2b6b43ee0.

kahrendt (approved):

PR #167, opened 2026-08-24
player role: carry the server's send-ahead on binary audio chunks

### LINKED ISSUE #156: Intermittent Wifi noise not measurable
Measuring `required_lead_time` and `min_buffer` accurately is still really challenging despite adding the `server_transmitted` field to stream/start and stream/clear. That gives a few data points, but the type of interruptions in low latency will almost never get captured by that info.

I was measuring my wifi environment (single family home, so not a dense apartment complex and relatively quiet RF wise) for some unrelated to Sendspin projects. I do have a doorbell camera that uses wifi, and on that channel is it will take over the air time 300ms+ at a time roughly every 5 minutes. We don't start and clear streams frequently enough to capture that, let alone capture it enough to assume it isn't an outlier. So on that wifi channel, all my measurements based on the stream/start ~100 ms or so, but in practice, every 5 minutes I will get a drop out. On another wifi channel (using my other AP), I have a neighbor's whose AP does a slow broadcast which interferes 100 to 200 ms every minute. This is the quieter channel and that signal is weak, but it will still cause disruptions. Again unless I get lucky and capture a stream/start or stream/clear at that window, I won't have any insight into it.

Two possible options that I can see: 
1) Use buffer underflows as my own single the buffer needs to increase. Easiest, but also the most misleading because not every underflow is due to network conditions (i.e., an OTA update check happens and slows us down for a long duration; that isn't typical and shouldnt' delay the latency for every client). 
2) Timestamp every binary audio chunk with the server's sent timestamp. I can then gather the statistics with a lot of data points and model it much better. 

We should also consider allowing or even mandating using Opus for low latency streaming because clients can use Opus's native packet loss concealment to smooth out a small disruption. That's a big ask though...\

kahrendt:
Players report `required_lead_time_ms` and `min_buffer_ms` but have no way to measure them. The only server transmit timestamps on the wire were the `server_transmitted` field on stream/start, stream/clear and stream/end, which sample the network once per stream event. Intermittent interference; e.g., a doorbell camera taking 300 ms of airtime every few minutes, almost never coincides with a stream event, so a player sizes its buffer from measurements that never saw the condition that underruns. This causes a too small `min_buffer_ms` causing frequent dropouts during live streams.

Audio chunks now carry `send_ahead`: the microseconds between the server transmitting the chunk and the chunk's playback timestamp. Recovering the transmit time from it gives a player a delay sample per chunk, enough to build a distribution rather than infer one from a handful of points.

`send_ahead` is a duration rather than a second absolute timestamp. It halves the added header (4 bytes, not 8) at a chunk rate where the overhead is worst in the low-latency case that motivates the change, and it cannot overflow in practice: uint32 microseconds covers about 71 minutes against a send-ahead bounded by buffer_capacity. The field saturates at both ends instead of wrapping, so a lead too large to represent degrades to a discarded sample rather than a plausible wrong one.

Defines where the transmit timestamp is taken, as late as the implementation permits, never at enqueue, and where the matching arrival time is taken. Without that, two conforming servers stamping at different points report different delays for the same network and players tune to whichever server they were developed against.

Drops `server_transmitted` from stream/end, which nothing consumes.

The artwork and visualizer roles are deliberately unchanged: artwork is never dropped for lateness, and a visualizer losing its first frames is cosmetic. The source role is unchanged too; its chunks already carry a capture timestamp, and splitting out the client's pipeline latency would give a server information it has no lever to act on.

Closes #156

## Breaking changes

**Binary audio chunk layout.** `send_ahead` occupies bytes 9-12, so the encoded audio frame moves from byte 9 to byte 13. Every player must update its chunk parser; one reading the payload at the old offset gets four bytes of header prepended to each frame and will fail to decode.

**`server_transmitted` stamping point is now normative.** [Transmit timestamps](messaging.md#transmit-timestamps) requires the value to be taken as late as the implementation permits and prohibits stamping at enqueue time. The field's previous definition said only "timestamp that the server transmitted this message", so a server stamping at enqueue was not clearly wrong before and is non-conforming now. This applies to `server/time`, `stream/start` and `stream/clear`; `server/time` is where it matters most, since the value feeds the client's time filter directly. `aiosendspin` already stamps immediately before sending and needs no change.

**`server_transmitted` removed from `stream/end`.** Nothing consumed it, but a client that treats the field as required will reject the message.


messaging.md
```diff
@@ -113,6 +113,14 @@ Each [`server/time`](#server--client-servertime) response provides the four time
 
 A player MUST NOT report `available: true` until its time filter has converged enough to begin scheduling playback. A source MUST NOT report `available: true` until its time filter has converged enough to timestamp captured audio.
 
+### Transmit timestamps
+
+Two things report when the server transmitted a message: the `server_transmitted` field, and the `send_ahead` interval carried in binary audio chunks. The server takes this time as late as its implementation permits, at the point the encoded bytes are handed to the transport for writing, after any application-level queueing or per-client scheduling. A server MUST NOT stamp a message at the time it is enqueued for later transmission.
```
maximmaxim345: > at the point the encoded bytes are handed to the transport for writing

The timestamp sits inside the message, so it has to be written before the bytes are encoded. How about: "immediately before encoding the message for transmission"?

roles/player/v1.md
```diff
@@ -67,6 +67,8 @@ State updates must be sent whenever any state changes, including when the volume
...
 
+**Measuring timing parameters:** A player derives `min_buffer_ms` from the distribution of arrival delay across audio chunks. For a chunk carrying [`send_ahead`](#server--client-audio-chunks-binary), that delay is `arrival - compute_client_time(timestamp - send_ahead)`, where `arrival` is the player's local receive time (see [Transmit timestamps](../../messaging.md#transmit-timestamps)) and `compute_client_time` is the time filter's server-to-local mapping. Players SHOULD size `min_buffer_ms` from the upper tail of the distribution, measured over a window long enough to include intermittent interference, and SHOULD discard samples taken before the time filter has converged. `required_lead_time_ms` is not derivable from this distribution alone: it is measured from a start trigger, and the chunks following a [`stream/start`](../../messaging.md#server--client-streamstart) that begins buffering from empty arrive under burst conditions that do not represent steady-state delay.
```
maximmaxim345: > For a chunk carrying `send_ahead`

Every chunk should carry it though, right?

maximmaxim345 (commented):
Left two nits.

This is also technically a breaking change if a server took `server_transmitted` at enqueue time for `server/time`. (I mean `aiosendspin` already took it right before sending, but still good to explicitly mention in the breaking changes section of this PR)

maximmaxim345 (approved):
LGTM.

Issue #134, opened 2026-08-12
Artwork transfers can starve audio: head-of-line blocking on the shared connection


arturpragacz:
## Warning / summary

Every message to a client shares one WebSocket, and the fragmentation rules make a fragmented message an *indivisible burst*: only one fragmented message may be in flight per direction, and the sender must finish it before sending **any** other frame. A Noise frame carries at most 65,518 bytes of payload, so any artwork image larger than ~64 KB becomes a contiguous fragment train that audio chunks cannot preempt. Even a server that wanted to interleave audio between fragments is forbidden from doing so by the spec.

The resulting audio stall is simply `image_size / link_throughput`, and it lands on top of whatever buffer cushion the player happens to hold at that moment. For buffered music mid-track the cushion is large and nothing audible happens. But:

- **If we want to target live audio**, players hold just `min_buffer_ms` of queued audio, since a live stream can never buffer further ahead. The whole point of the low-latency design is that this can be small (tens of ms on a good network), and then a perfectly ordinary track-change JPEG is enough to underrun every player that shares the connection with an artwork channel.
- **Even without live audio**, the spec-mandated BMP support produces megabyte-class images that threaten conservative buffers too. An 800×480 24-bit BMP is ~1.1 MB, which is ~450 ms of wire occupancy at 20 Mbps. Devices that request BMP (cheap decode, no image library) are exactly the constrained Wi-Fi devices with 10–25 Mbps of real throughput. A display-speaker asking for BMP artwork is a dropout waiting to happen, at today's default buffers.

## Numbers

Stall = image size / effective link throughput; one image, one channel:

| Image | Size | @100 Mbps | @20 Mbps |
|---|---|---|---|
| 480×480 JPEG | ~40 KB | 3 ms | 16 ms |
| 1000×1000 JPEG/PNG | 150–400 KB | 12–30 ms | 60–160 ms |
| 800×480 BMP (24bpp) | ~1.1 MB | ~90 ms | ~450 ms |
| Track change: album + artist, 2×1000×1000 JPEG/PNG | 300–800 KB | 24–64 ms | 120–320 ms |
| Track change: 2 BMP channels, 800×480 | ~2.3 MB | ~180 ms | ~0.9 s |

[...]

PR #168, opened 2026-08-24
Letterbox art and remove BMP support

### LINKED ISSUE #128: Artwork dimensions letterboxing ambiguity

kahrendt:
The previous artwork scaling rule said the server "will scale images to fit within the specified dimensions while preserving aspect ratio", which reads as fit-inside: a square cover requested at 800x480 would arrive 480x480. The spec will adopt what `aiosendspin` already does: letterbox.

Changes:

- `width`/`height` are the dimensions of the delivered image, not upper bounds (renamed from `media_width` and `media_height`)
- The server MUST pad to those dimensions with black and MUST NOT crop
- `stream/start` `width`/`height` MUST equal the declared `width`/`height` rather than merely not exceeding them
- Drop `'bmp'` from the artwork formats, leaving `'jpeg'` and `'png'` (see comments in #134)

The equality rule settles the case that blocked scheduled artwork updates: with the delivered size pinned to the declaration, a later image cannot arrive at a different aspect ratio than the resolution `stream/start` announced.

I'll make a future PR splitting large images across messages, the remaining part of #134.

Closes #128

## Breaking Changes
- BMP is now unsupported
- `media_width`/`media_height` are renamed to `width`/`height`
- Servers must letterbox artwork

### DIFF
--- a/roles/artwork/v1.md
+++ b/roles/artwork/v1.md
@@ -11,10 +11,10 @@ The `artwork@v1_support` object in [`client/hello`](../../messaging.md#client--s
   - `channels`: object[] - list of supported artwork channels (length 1-4), array index is the channel number
     - `source`: 'album' | 'artist' | 'none' - artwork source type
     - `format`: 'jpeg' | 'png' | 'bmp' - image format identifier
-    - `media_width`: integer - max width in pixels
-    - `media_height`: integer - max height in pixels
+    - `media_width`: integer - width in pixels of the delivered image
+    - `media_height`: integer - height in pixels of the delivered image
 
-The server MUST scale images to fit within the specified dimensions while preserving aspect ratio.
+The server MUST deliver each image at exactly the `media_width` by `media_height` declared for the channel. It MUST scale the source image to fit within those dimensions preserving its aspect ratio, MUST pad the remaining area with black, and MUST NOT crop the image.
 
 **Note:** Clients can support 1-4 independent artwork channels depending on their display capabilities. The channel number is determined by array position: `channels[0]` is channel 0 (binary message type 8), `channels[1]` is channel 1 (binary message type 9), etc.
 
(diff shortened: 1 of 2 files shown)


roles/artwork/v1.md
```diff
@@ -44,13 +44,13 @@ The `artwork` object in [`stream/start`](../../messaging.md#server--client-strea
...
     - `height`: integer - height in pixels of the encoded image
 
 The `channels` array covers every channel index the client declared in [`artwork@v1_support`](#client--server-clienthello-artworkv1-support-object) in the same order. A channel the server is not streaming is represented as `source: 'none'`.
 
-Each channel's configuration MUST stay within the client's current capability for that channel: the [`client/hello`](#client--server-clienthello-artworkv1-support-object) declaration, as later modified by [`stream/request-format`](#client--server-streamrequest-format-artwork-object). The `source` and `format` MUST match the declaration, and `width`/`height` MUST NOT exceed the declared `media_width`/`media_height`.
+Each channel's configuration MUST match the client's current capability for that channel: the [`client/hello`](#client--server-clienthello-artworkv1-support-object) declaration, as later modified by [`stream/request-format`](#client--server-streamrequest-format-artwork-object). The `source` and `format` MUST match the declaration, and `width`/`height` MUST equal the declared `media_width`/`media_height`.
```
maximmaxim345: Very minor: does "as later modified by stream/request-format" include a request the server declined? I read it as only the honored ones, but might be worth writing it as: "later modified by any stream/request-format the server honored".

maximmaxim345 (commented):
Can you also rename `media_width` and `media_height` to just `width` and `height` in this PR? See: https://github.com/Sendspin/spec/issues/128#issuecomment-5346309284

maximmaxim345 (approved):
Thanks!

PR #169, opened 2026-08-24
Specify client reset handling

arturpragacz:
Specify client reset handling:
- Add Sentinel PSK fallback to improve handling by server.
- Clarify what state should be cleared.


maximmaxim345 (approved):
Added a link to the new Factory Reset definition in the "Sentinel Fallback" section, but looks good other than that.

Thanks @arturpragacz !

PR #170, opened 2026-08-24
Shape unpaired access UX

### LINKED ISSUE #157: Simplify unpaired access UX for using Sendspin
- Require UI action to allow unpaired access (even just pressing play should be enough)
- Allow source role without pairing, but say that severs need to ask user if the stream should be accepted (implicitly solves the user friction mentioned in #151)

arturpragacz:
Shape unpaired access UX.

Resolves #157.


roles/source/v1.md
```diff
@@ -7,7 +7,7 @@ A device MAY implement both the `source` and `player` roles (e.g., a speaker wit
...
 
-**Pairing required:** A source captures potentially sensitive audio (microphone, line-in), so `source@v1` MUST only run on a paired connection ([trust level](../../README.md#definitions) `user`) and a source client MUST NOT stream when the trust level is `none`. If a server activates `source@v1` at trust level `none`, the client refuses it and closes the connection, following the central rules in [`server/activate`](../../messaging.md#server--client-serveractivate).
+**Unpaired access:** As `source@v1` exposes the server's audio stack to arbitrary input, activating it for an unauthenticated device requires explicit [approval](../../pairing.md#unpaired-access). Clients with a privacy-sensitive input such as a microphone SHOULD ship with unpaired access disabled: the captured audio would otherwise be accessible to anyone on the network.
```
maximmaxim345: Very minor nitpick, feel free to dismiss if you think this doesn't need to be addressed:

Only disadvantage of this change I can see if a player shipped with a privacy-sensitive source role: With the previous version it can still enable unpaired_access while protecting the source role. After this PR, it must/should lock down playback without pairing completely.

But I imagine only very few cases actually fall into the "privacy-sensitive" case so this is just a theoretical nitpick.

maximmaxim345: Small inconsistency I think: this section says explicit approval, while "Pairing > Unpaired Access" allows implied approval from an operator action that clearly means to use the client.

I think we decided on a "two-stage" intent: (implied approval activates the role, but the server holds the command: "start" until the operator explicitly accepts), could we spell that out here? Since a source MUST NOT stream before start, it's worth saying that nothing leaves the client until the operator says yes.

arturpragacz: Yes, that's an inherent limitation of this approach. We could add something like `unpaired_roles`, where the client could say which roles can be used unpaired and which cannot, but maybe that's overkill, at least for Sendspin 1.0?

maximmaxim345 (approved):
Looks good now, thanks!

PR #173, opened 2026-08-25
Deliver spoken pairing-code digits as server-supplied audio pack

arturpragacz:
Deliver spoken pairing-code digits as server-supplied audio pack.

Speaker clients no longer need bundled voice assets. In the dynamic audio pairing flow the server supplies a pack of ten short clips, one per digit, in a language of its choosing, and the client plays them in pairing code order.

- Client advertises `digit_audio` in its `dynamic_pairing_code` descriptor: accepted `formats` (codec, sample rate, bit depth) and `max_bytes`.
- Server picks the format, announces it as `digit_audio` in the pairing `server/activate`.
- New binary message type `2`: one clip per message, ten in ascending order. Clips are mono, at most 2 seconds each.
- `languages` moves from the pairing activation to `server/hello` as a general operator-language hint.

Needs #172 merged first.


messaging.md
```diff
@@ -200,8 +202,12 @@ Only after receiving the initial `server/activate` should the client send any ot
...
+  - `format?`: 'digits' | 'qr_code' - the dynamic [emission format](pairing.md#dynamic-pairing-code-flow), drawn from the client's `dynamic_pairing_code` descriptor. Required when `method` is `'dynamic_pairing_code'`; absent otherwise. The server selects `qr_code` only when its operator interface can scan a QR code, and - for a client whose descriptor carries `digit_audio` - `digits` only when it can produce one of its `formats` within its `max_bytes`.
+  - `digit_audio?`: object - announces server-supplied [digit audio](pairing.md#dynamic-pairing-code-flow) clips for this attempt, in one of the `digit_audio.formats` from the client's descriptor. Required when `format` is `'digits'` and the client's descriptor carries `digit_audio`; absent otherwise.
+    - `codec`: 'opus' | 'flac' | 'pcm'
+    - `sample_rate`: integer - sample rate in Hz
```
kahrendt: Opus only supports 48 kHz

pairing.md
```diff
@@ -381,3 +394,19 @@ Aborts a pairing attempt, started or not. With reason `concurrent_attempt` the s
...
+
+One clip of a [digit audio pack](#dynamic-pairing-code-flow).
+
+Sent only in an attempt whose [`server/activate`](messaging.md#server--client-serveractivate) announced `digit_audio`, after [`client/pair-init`](#client--server-clientpair-init) and before [`server/pair-init`](#server--client-serverpair-init): ten messages in ascending digit order, together at most the descriptor's [`max_bytes`](#client--server-clienthello-pair-method-descriptor). A clip outside that window, out of order, duplicated, or with a digit above 9 is a [sequence violation](#messages).
+
+- Byte 0: message type `2` (uint8)
+- Byte 1: digit (uint8) - the decimal digit `0`-`9` the clip speaks
+- Rest of bytes: the clip in the announced format
+
+**Clip contents:** Each clip carries one whole spoken digit - mono, at most 2 seconds - in the format announced in [`server/activate`](messaging.md#server--client-serveractivate); a clip whose embedded stream parameters contradict it - a FLAC STREAMINFO with other channels, sample rate, or bit depth, an OpusHead with other channels - is malformed. Per codec:
```
kahrendt: We don't ever send an `OpusHead` packet as we don't have the Ogg framing. This gets a bit tricky actually because we've only ever sent full Opus packets before one-by-one. If we want to frame it to make it standalone, then we'd need an Ogg Opus file. I believe Opus also has a self-limiting format for just raw packets, but I've never used it. I'm fine with using Ogg Opus files, that's the easiest way to support it.

arturpragacz: That's what I intended actually: https://github.com/Sendspin/spec/pull/173/changes#diff-42c453559257720e9e678b17d5fac9329792c4b11466c94f8e16be481d46875dR412

messaging.md
```diff
@@ -200,8 +202,12 @@ Only after receiving the initial `server/activate` should the client send any ot
...
+    - `bit_depth`: integer - bit depth; ignored for `opus`
+    - `codec_header?`: string - codec header encoded as standard Base64 ([RFC 4648 section 4](https://www.rfc-editor.org/rfc/rfc4648#section-4), padding included). Required for `flac`, absent otherwise
```
kahrendt: If we are sending standalone files, the flac file will have its own embedded header instead of putting it here.

kahrendt: ~Wait why is the client sending this in the first place?~ Edit: misread

pairing.md
```diff
@@ -289,20 +296,26 @@ with `label` `"sendspin-pair-psk-wrap-v1"` for `wrapped_psk` and `"sendspin-pair
...
 
 - `method`: 'dynamic_pairing_code' | 'pairing_psk' | 'static_pairing_code' - the pairing method identifier.
-- `out_channels?`: ('display' | 'speaker')[] - informational hint for `dynamic_pairing_code` only, listing the channels through which the per-session pairing code is conveyed to the operator.
+- `out_channels?`: ('display' | 'speaker')[] - the channels through which the per-session pairing code is conveyed to the operator. Required on `dynamic_pairing_code` descriptors, absent on others.
 - `formats?`: ('digits' | 'qr_code')[] - the [emission formats](#dynamic-pairing-code-flow) the client offers. Required on `dynamic_pairing_code` descriptors, absent on others. Non-empty; `qr_code` requires a display able to render a QR code.
+- `digit_audio?`: object - the server-supplied [digit audio pack](#dynamic-pairing-code-flow) the client wants. Required on `dynamic_pairing_code` descriptors whose `out_channels` include `'speaker'`; absent otherwise.
+  - `formats`: object[] - formats the client accepts, in descending preference
+    - `codec`: 'opus' | 'flac' | 'pcm' - codec identifier
+    - `sample_rate`: integer - sample rate in Hz (e.g., 16000)
```
kahrendt: Here's where we need to note opus is only 48 kHz

messaging.md
kahrendt: Wrong spot for this

pairing.md
```diff
@@ -289,20 +296,26 @@ with `label` `"sendspin-pair-psk-wrap-v1"` for `wrapped_psk` and `"sendspin-pair
...
 
 - `method`: 'dynamic_pairing_code' | 'pairing_psk' | 'static_pairing_code' - the pairing method identifier.
-- `out_channels?`: ('display' | 'speaker')[] - informational hint for `dynamic_pairing_code` only, listing the channels through which the per-session pairing code is conveyed to the operator.
+- `out_channels?`: ('display' | 'speaker')[] - the channels through which the per-session pairing code is conveyed to the operator. Required on `dynamic_pairing_code` descriptors, absent on others.
 - `formats?`: ('digits' | 'qr_code')[] - the [emission formats](#dynamic-pairing-code-flow) the client offers. Required on `dynamic_pairing_code` descriptors, absent on others. Non-empty; `qr_code` requires a display able to render a QR code.
+- `digit_audio?`: object - the server-supplied [digit audio pack](#dynamic-pairing-code-flow) the client wants. Required on `dynamic_pairing_code` descriptors whose `out_channels` include `'speaker'`; absent otherwise.
+  - `formats`: object[] - formats the client accepts, in descending preference
```
maximmaxim345: `formats` needs to have a non-empty requirement, otherwise this could cause undefined behavior.

pairing.md
```diff
@@ -141,7 +144,11 @@ qr_code:  code     = digest[0..23]
...
...ver-clienthello-pair-method-descriptor) - the `formats` it accepts, in descending preference, and `max_bytes`, the largest encoded pack size it accepts. The server picks the highest-priority format it can produce within `max_bytes` - servers MUST support all three codecs - and announces it as `digit_audio` in the pairing [`server/activate`](messaging.md#server--client-serveractivate). After [`client/pair-init`](#client--server-clientpair-init) it delivers the clips as [digit audio clip](#server--client-digit-audio-clip-binary) messages in ascending digit order, each at most 2 seconds of audio and together at most `max_bytes` encoded. [`server/pair-init`](#server--client-serverpair-init) then completes the pack. The client emits by playing the clip for each code digit in turn. The pack is presentation-only - clips never enter the derivation or `PRS` - and is discarded when the attempt ends. A `digit_audio` naming a format the client did not advertise, a clip over 2 seconds, a pack over `max_bytes`, a clip undecodable or whose embedded stream parameters contradict the announced format, or a pack still incomplete when `server/pair-init` arrives is a [protocol error](#protocol-errors).
```
maximmaxim345: Hmm, I think there is a potential for abuse. Was thinking about the 2 second limit that can maybe even fits fast and short sentences:
- A rogue server does the handshakes
- since the client isn't escalated it also isn't gesture gated
- server sends 10 identical audio samples
- client derives the code and plays one per digit
- server drops the connection and restarts from step one

You essentially get an endless rapid fire of 2 seconds of audio since the failure counter never gets incremented.

But this was also the case before this PR (though without the custom audio part).

messaging.md
```diff
@@ -200,8 +202,12 @@ Only after receiving the initial `server/activate` should the client send any ot
...
-  - `format?`: 'digits' | 'qr_code' - the dynamic [emission format](pairing.md#dynamic-pairing-code-flow), drawn from the client's `dynamic_pairing_code` descriptor. Required when `method` is `'dynamic_pairing_code'`; absent otherwise. The server selects `qr_code` only when its operator interface can scan a QR code.
-  - `languages?`: string[] - non-empty list of [BCP 47](https://www.rfc-editor.org/rfc/rfc5646) language tags in descending operator preference (e.g. `["ca", "es", "en"]`), for spoken [pairing code emission](pairing.md#dynamic-pairing-code-flow). Optional when `method` is `'dynamic_pairing_code'` with the `digits` emission format; absent otherwise.
+  - `format?`: 'digits' | 'qr_code' - the dynamic [emission format](pairing.md#dynamic-pairing-code-flow), drawn from the client's `dynamic_pairing_code` descriptor. Required when `method` is `'dynamic_pairing_code'`; absent otherwise. The server selects `qr_code` only when its operator interface can scan a QR code, and - for a client whose descriptor carries `digit_audio` - `digits` only when it can produce one of its `formats` within its `max_bytes`.
```
maximmaxim345: Nit: Spelling out `digit_audio.formats` and `digit_audio.max_bytes` makes it a bit easier to understand.

pairing.md
```diff
@@ -289,20 +296,26 @@ with `label` `"sendspin-pair-psk-wrap-v1"` for `wrapped_psk` and `"sendspin-pair
...
+- `out_channels?`: ('display' | 'speaker')[] - the channels through which the per-session pairing code is conveyed to the operator. Required on `dynamic_pairing_code` descriptors, absent on others.
 - `formats?`: ('digits' | 'qr_code')[] - the [emission formats](#dynamic-pairing-code-flow) the client offers. Required on `dynamic_pairing_code` descriptors, absent on others. Non-empty; `qr_code` requires a display able to render a QR code.
+- `digit_audio?`: object - the server-supplied [digit audio pack](#dynamic-pairing-code-flow) the client wants. Required on `dynamic_pairing_code` descriptors whose `out_channels` include `'speaker'`; absent otherwise.
+  - `formats`: object[] - formats the client accepts, in descending preference
+    - `codec`: 'opus' | 'flac' | 'pcm' - codec identifier
+    - `sample_rate`: integer - sample rate in Hz (e.g., 16000)
+    - `bit_depth`: integer - bit depth (e.g., 16); meaningful for `pcm` and `flac` only, ignored for `opus`
+  - `max_bytes`: integer - maximum total encoded size of the ten clips in bytes
```
maximmaxim345, line 312: Idea: Should we give a minimum size to not land in the situation that some clients can't be paired with some servers/languages?

pairing.md
arturpragacz: Hmmm, I'm not an expert when it comes to Opus, but here's what Claude claims:

> Opus is not limited to 48 kHz on the API side. Encoder and decoder both take 8, 12, 16, 24 or 48 kHz (RFC 6716 section 2; libopus takes Fs when creating the decoder). 48 kHz is only the codec's internal rate and the Ogg granule clock, the decoder resamples to whatever you ask for. The player role treats `sample_rate` for opus the same way already.

Happy to adjust it anyway, I just wanted to confirm that's what we want.

pairing.md
arturpragacz: I'm aware, that's why I limited the clips to 2 seconds, I don't think you can fit much in that.

It's not a security issue as nothing is compromised, it just adds a little susceptibility to nuisance, which was there to a similar extent before as the device could already be made to speak a code unprompted before this PR. The new part is only that the server gets to pick the audio.

I don't think it would be sensible to gate every attempt on a gesture for it. That worsens the normal pairing experience to cover an edge case. Anyone who needs to deploy Sendspin in a really hostile network, where preventing that is really required, can already disable the dynamic method entirely.

Also this whole thing is obviously completely moot if `unpaired_access` is enabled, as then anyone can play anything.

That being said, I decided to add a small mitigation: the failure counter now increments when the client starts emitting a code instead of on a failed confirmation. That should help with the most egregious cases.

pairing.md
arturpragacz: Maybe that wouldn't be a bad idea.

@kahrendt What do you think we could put here? Would something like 32 KiB be reasonable?

It would be easier to make this decision if we first had recordings for a bunch of languages that we want to support.

pairing.md
maximmaxim345: Have no better alternative ideas than incrementing the failure counter as well.
I think "an abandoned attempt is not an inner-authentication failure" also needs to be updated now.

For Music Assistant we also probably need to add another step explaining that it will speak the code after pressing OK. So we minimize the wasted attempts.

pairing.md
kahrendt: All audio is encoded at 48 kHz. If a client wants to output it at 16 kHz, it just tells the opus decoder to that and it resamples (very poorly). That decoder setting doesn't change anything regardless if the encoder used a 16 kHz or 48 kHz input source.

kahrendt: So I guess along those lines, it doesn't actually matter the sample rate cause the encoder will handle it... so then I guess it doesn't matter what we allow here (though if we want to be careful, we wouldn't allow 44.1 kHz if we want to follow the Opus RFC spec)

pairing.md
kahrendt: 32 KiB is very restrictive for PCM. That's only 1 second at 16 kHz mono audio at 16 bits per sample. I would imagine most clients would just use their output format sample rate; i.e., 48 kHz, and that is only 0.33 seconds.

arturpragacz: Sure, but should be enough for Opus probably?

Otherwise we would have to mandate it based on the supported codecs, so that gets a little more complicated.

I'm leaning towards just leaving it out for now. We can reevaluate once we have the recordings.

pairing.md
arturpragacz: > an abandoned attempt is not an inner-authentication failure

Well, it still isn't. The failure counter now just doesn't increase based on authentical failures.

Also if the attempt is abandoned before the pin is emitted, it also doesn't increase the counter.

pairing.md
maximmaxim345: What is the benefit of supporting multiple formats for audio packs?
The server in either way will have to support any samplerate/bitdepth/codec (so you can actually pair it successfully).
That will also simplify the `max_bytes` recommendations in case we define different ones per codec.

pairing.md
kahrendt: Since it doesn't matter on either side, let's not even mention it

pairing.md
arturpragacz: I removed the digit audio format negotiation.

maximmaxim345 (approved):
I think this looks good now.

@arturpragacz can you create a issue (and add it to the project board) so we don't forget to add `max_bytes` recommendations/requirements?

arturpragacz:
I added it here: https://github.com/Sendspin/backlog/issues/41.

PR #175, opened 2026-08-26
Send full state instead of deltas

### LINKED ISSUE #149: Define one shared delta-merge rule for `client/state`, `server/state`, and `group/update`
_Revised after the review by @maximmaxim345. The proposal is now smaller: it unifies the mechanical rules, and it keeps the difference in the minimum content._

## Problem

Three messages carry a state update: `client/state`, `server/state` and `group/update`. Each message gives its own merge rules.

| | `client/state` | `server/state` | `group/update` |
|---|---|---|---|
| Merge | the server MUST merge, and keeps absent fields | the client "will merge" | the client "should merge" |
| `null` clears | not defined | a null leaf clears the field, a null role object clears the role | not defined |
| Merge depth | not defined | shallow, a nested object is replaced in full | not defined |
| First full message | the first message, and the full role state when a role becomes active | the first message for each role, and the first message after the role becomes active again | the first message on a connection |
| Minimum content | MAY send only changed fields, MAY send the full state | "Only include fields that have changed" | "only the changed fields" |

The first four rows are the same procedure with three descriptions. Only the last row is a true difference.

The spec says that the difference is intentional: "The asymmetry with `client/state` is deliberate". But it does not give the reason.

## Change

Give the mechanical rules one time, in the Communication section of messaging.md. Keep the difference in the minimum content, and give the reason for it.

> State-carrying messages (`client/state`, `server/state`, `group/update`) are delta updates. The first message for a scope MUST carry that scope's full state: the initial `client/state` on a connection and a role's full state when it becomes active in `active_roles`; the first `server/state` for a role on a connection and the first after the role is re-added to `active_roles`; the first `group/update` on a connection. Subsequent messages carry the fields that have changed; the receiver MUST merge each message into existing state, retaining the last value of any absent field. The merge is shallow: a nested object is replaced or cleared as a whole, never deep-merged, so nested objects are always sent complete. Where a field or role object is clearable (typed `| null`), setting it to `null` clears it; a role object set to `null` clears all of that role's state. A role's definition MAY defer when a merge is applied (see scheduled updates).
>
> Servers SHOULD send only the fields that have changed. Clients MAY instead resend unchanged fields, up to their full state: the server is the more capable side, so a client may keep its implementation simple.

Then each message keeps one sentence and its field list. Remove the note about the intentional difference, because the new rule gives the reason.

The note in `server/activate` and the note in `roles/player/v1.md` already point to these rules. Change the anchors only.

## Effect on behavior

None. The rule keeps the current minimum-content requirement for each direction.

## Open point: the metadata timestamp

@maximmaxim345 found a problem in the first version of this proposal. It permitted the server to resend unchanged fields. That is not safe for the `metadata` role.

A metadata client calculates the position from the `timestamp` and the `progress` of the last message with a `progress` object. Thus an unchanged `progress` under a new `timestamp` moves the calculated position.

The revised rule prevents this, because the server keeps the minimum-content requirement. But PR #135 makes the server resend the fields of a scheduled message. Thus the two rules meet in that case.

Select one solution:

1. A resent value keeps its original timestamp.
2. The `metadata` role keeps its own rule, because its `timestamp` gives the fields a different function.

Solution 2 is possibly better, because the problem comes from the role, not from the merge procedure.

## Correction to make at the same time

The controller object in `server/state` has mandatory leaves without the type `| null`. Thus the general null rule and the schema do not fully agree today. The new rule must say that the field type controls this behavior.

## Sequence with PR #135

PR #135 changes the same `server/state` paragraphs. It adds scheduled updates: a future `timestamp` delays the merge, a null role object removes a pending update, and the first message for a role must have a past or present timestamp.

These rules apply to one role. They are not merge rules. Thus they stay with `server/state` and the role files, and they use the sentence "A role's definition MAY defer when a merge is applied".

Merge PR #135 first. Then apply this change on top of it.

kahrendt:
`client/state`, `server/state` and `group/update` each defined their own merge rules, and the three disagreed on whether merging was required, whether `null` cleared a field, how deep the merge went, and which message had to be complete (#149). Rather than reconcile them into one rule, drop merging.

## Changes:

- All three messages carry full state on every message.
- For `client/state` and `server/state` the role object is the unit: an included object carries that role's full state, and an omitted one leaves that role's state unchanged. Differential updates survive at role granularity, which is what the optional role objects were always for.
- Drop the `| null` variant from leaf fields in the `metadata` and `color` objects. A role object may still be `null` in `server/state`, which is how the server ends a role's state when it drops the role from `active_roles`, and it was never a delta mechanism.
- `group/update`'s `playback_state`, `group_id` and `group_name` are no longer optional.

What this costs is bandwidth on metadata text, which is sent on state change rather than continuously; anything large already travels as a binary message. What it saves is the reconciliation layer on the receiving side, which has been a steady source of bugs and runs to a few hundred lines in a client that gets it right.

## Fields whose absence carried a second meaning

Dropping the merge silently repurposes every optional field that used to mean "unchanged". Each one is now stated where the reader meets the field, rather than in a rule they have to remember:

- `metadata.progress`: omitting it clears the client's position, so the server includes it in every `metadata` state that has a position to report. Track position is calculated from the current metadata state rather than from the last message that happened to include `progress`, which only made sense while absent meant unchanged. A resent `progress` is measured at the new `timestamp` rather than copied forward under it, which the merge rules used to make moot by never resending it at all.
- The player state object's `supported_commands`: absence now says the player accepts no commands, which is what `[]` already says. The `?` goes rather than the redundant encoding staying around to silently revoke `set_output_delay`. Both other `supported_commands` in the spec are already non-optional.
- Role objects themselves, covered by the omitted-object rule above.

## Follow-on cleanups

- The player role's "Delta updates" paragraph is gone.
- `REQUIRED for players` on the three non-optional fields in the player state object goes with it. It was an anchor for that paragraph, `?` already marks optionality (see the note under Communication), and the object is player-only.
- The `available: false` handling no longer describes sending a `group/update` carrying only `playback_state`.

This conflicts with #135 on `messaging.md`, `roles/metadata/v1.md` and `roles/color/v1.md`, in that PR's favour: with no merging, its rule that the first message after a scheduled one must repeat the scheduled message's fields has nothing left to do.

Out of scope: the `management/set-pairing-config` patch semantics in `management.md`, which is a configuration API rather than state streaming.

Closes #149
Refs #101

## Breaking changes

Both directions break, not just the server.

Servers must send full `server/state` and `group/update` messages instead of deltas. A client that still merges keeps stale leaves, because clearing a leaf is now omission rather than an explicit `null`.

Clients must send full `client/state` messages, and a player must now always carry `supported_commands`. A client that still sends deltas has them read as full state, so every field it omits is dropped rather than retained. Neither side errors.


roles/metadata/v1.md
```diff
@@ -7,21 +7,21 @@ The `metadata` object in [`server/state`](../../messaging.md#server--client-serv
...
-  - `track?`: integer | null - track number on the album (1-indexed), null if unknown or not applicable
-  - `progress?`: object | null - playback progress information. The server must send this object whenever playback state changes (play, pause, resume, seek, playback speed change)
+  - `title?`: string - track title
+  - `artist?`: string - primary artist(s)
+  - `album_artist?`: string - album artist(s)
+  - `album?`: string - name of the album or release that this track belongs to
+  - `artwork_url?`: string - URL to artwork image. Useful for clients that want to forward metadata to external systems or for powerful clients that can fetch and process images themselves
+  - `year?`: integer - release year in YYYY format
+  - `track?`: integer - track number on the album (1-indexed), absent if unknown or not applicable
+  - `progress?`: object - playback progress information. The server must send this object whenever playback state changes (play, pause, resume, seek, playback speed change)
```
maximmaxim345:
```suggestion
  - `progress?`: object - playback progress information. Omitting it clears the client's position, so include it in every `metadata` state that has a position to report. The server must send a new `metadata` state whenever playback state changes (play, pause, resume, seek, playback speed change)
```

The wording still sounded like you should only send the progress object (which of course clears all other metadata).

messaging.md
```diff
@@ -314,13 +316,9 @@ Client sends commands to the server. Contains command objects based on the clien
 
 Server sends state updates to the client. Contains role-specific state objects.
 
-Only include fields that have changed. The client will merge these updates into existing state. A leaf field set to `null` should be cleared from the client's state; a whole role object set to `null` clears all of that role's state.
-
-The merge is shallow: a nested object (e.g., `metadata.progress`) is replaced or cleared as a whole, never deep-merged, so nested objects are always sent complete.
+Every message MUST carry the full state of each role object it includes.
```
maximmaxim345: While omitting a field is still allowed (the `?` in `color?` for example), that case is now undefined behavior.

maximmaxim345 (commented):
Can you also expand the breaking changes section of the description to clients that used differential updates? `client/state` also changes after all.

messaging.md
maximmaxim345: I think we should still allow differential updates on a role by role basis. I assume that also was your intention since you kept the `?` and `null` here and in the other messages.

maximmaxim345 (approved):

PR #176, opened 2026-08-27
Clamp out-of-range `output_delay_ms`

### LINKED ISSUE #93: Static delay: specify what a client does with an out-of-range `static_delay_ms`

maximmaxim345:
The spec documents the `output_delay_ms` range as 0-5000 in both the `client/state` player object and the `set_output_delay` parameters, but never says what a client does with a value outside it, so SDK behavior diverges: some obey the range, some clamp, and one flips a negative value's sign, which is audible. This adds one sentence to the "Output delay" paragraph requiring clients to clamp the value into 0-5000, so a bad value degrades to the nearest valid delay instead of disrupting playback. The issue also proposed a SHOULD-log clause, left out because the spec has no logging requirements anywhere and logging is neither observable on the wire nor testable by a conformance suite.

Closes #93


kahrendt (approved):
Thanks!

PR #177, opened 2026-08-27
Allow players to change `supported_commands` mid-session

### LINKED ISSUE #142: player@v1: consolidate `supported_commands` into `client/state` and give the timing/buffering server rules their own section

maximmaxim345:
`supported_commands` was split across two messages, `'volume'` and `'mute'` in the `client/hello` support object and `'set_output_delay'` in `client/state`, so a player could not change which commands it accepts without reconnecting and `server/command` had to name both locations as sources of truth. The whole list moves into the `client/state` player object.

The split also failed silently: an SDK reading only `client/hello` ignores `set_output_delay`, one reading only `client/state` ignores `volume` and `mute`, and neither errors. Consolidating matches the CONTRIBUTING rule that a field expected to change during a connection belongs in `client/state`.

This breaks `player@v1` on the wire, and a server can no longer see command capabilities at `server/activate` time, which is exotic but worth stating.

#175 already made `supported_commands` non-optional and defined an empty list as accepting no commands, so the field line itself only gains `'volume'` and `'mute'`, and the `client/hello` copy is deleted.

# Rules the move requires

A server MUST NOT send a player command before the client has sent a `client/state` containing the `player` object, mirroring the existing rule for binary data. The move opens that window because `server/activate` precedes the first state message, and the player object is the unit that carries `supported_commands`.

`supported_commands` advertises settability, not reportability. A player MAY report `volume` or `muted` without accepting the matching command, an amplifier with a physical volume knob being the real case, and capability cannot be inferred from field presence because `output_delay_ms` is never optional either way. A server MUST NOT treat a reported `volume` or `muted` as settable while the matching command is absent, though it MAY still surface the value read-only.

Group volume averages over volume-capable players, so when the last capable player withdraws the reported value jumps to 100 and the control disappears with nobody having touched a volume. Both reading rules now say so and require a recompute when a player's support changes.

# Group volume writes

The group volume algorithm already disagreed with itself before this change: it computed its delta against the average of all player volumes and sent a command to every player, while the reading rules excluded players without volume support. The move makes the mismatch reachable mid-session, so both paths now filter on capability, and setting group mute gets the same fix. This part is separable if you would rather it went on its own.

# Editorial

The "Server behavior" bullets sat under the `client/hello` support-object heading while describing `required_lead_time_ms` and `min_buffer_ms`, which live in `client/state`, and two cross-references pointed at that mismatched anchor. The text merges into "Server Audio Send Constraints", which already carries the overlapping lead and send-ahead rules.

Closes #142


arturpragacz (commented):
Weren't we supposed to align all dynamically changing values and move them all to `client/state`? So including for example `visualizer`'s and `artwork`'s `stream/request-format` fields.

maximmaxim345:
> Weren't we supposed to align all dynamically changing values and move them all to `client/state`? So including for example `visualizer`'s and `artwork`'s `stream/request-format` fields.

Thats for a different PR though since it touches multiple roles, I'll write that once #175 is merged.

roles/controller/v1.md
```diff
@@ -46,11 +46,11 @@ Control the group that's playing and switch groups. Only valid from clients with
      - All delta has been successfully applied, or
      - All players are clamped at their volume boundaries
 
-The loop only computes the final per-player volumes; once it completes, the server sends each player a single volume command — intermediate values are never sent.
+The loop only computes the final per-player volumes; once it completes, the server sends each of those players a single volume command — intermediate values are never sent.
```
kahrendt: The first 2 steps specify players supporting the `volume` command. Steps 3 and 4 are not as specific (though 3 is probably fine because its each of those players' inheriting the specifications in steps 1 and 2). Maybe just start with defining a term for a set of players that are eligible before the algorithm starts and just use that term throughout.

roles/controller/v1.md
```diff
@@ -77,6 +77,6 @@ The `controller` object in [`server/state`](../../messaging.md#server--client-se
...
 
-**Reading group volume:** Group volume is the average of the volumes of players in the group that support the `volume` command. Players without volume support are excluded from the calculation. If no player in the group supports `volume`, group volume is reported as 100 and `'volume'` is dropped from the controller `supported_commands`.
+**Reading group volume:** Group volume is the average of the volumes of players in the group that support the `volume` command. Players without volume support are excluded from the calculation. If no player in the group supports `volume`, group volume is reported as 100 and `'volume'` is dropped from the controller `supported_commands`. A player MAY change its [`supported_commands`](../player/v1.md#client--server-clientstate-player-object) mid-session, so the server recomputes both the group volume and the controller `supported_commands` whenever a player's volume support changes. When the last volume-capable player withdraws support, the reported group volume jumps to 100 and the group volume control disappears without anyone having changed a volume.
```
kahrendt: "When the last volume-capable player withdraws support, the reported group volume jumps to 100 and the group volume control disappears without anyone having changed a volume." 
Reads a bit off for the spec, especially the "without anyone having change a volume" part. Could you rewrite it to be more normative?

roles/player/v1.md
```diff
@@ -90,11 +80,13 @@ The `player` object in [`server/command`](../../messaging.md#server--client-serv
 Request the player to perform an action, e.g., change volume or mute state.
 
 - `player`: object
-  - `command`: 'volume' | 'mute' | 'set_output_delay' - must be listed in `supported_commands` from [`player@v1_support`](#client--server-clienthello-playerv1-support-object) or from [`client/state`](#client--server-clientstate-player-object); unlisted commands are ignored by the client
+  - `command`: 'volume' | 'mute' | 'set_output_delay' - must be listed in `supported_commands` from [`client/state`](#client--server-clientstate-player-object); unlisted commands are ignored by the client
   - `volume?`: integer - volume range 0-100, only set if `command` is `volume`
   - `mute?`: boolean - true to mute, false to unmute, only set if `command` is `mute`
   - `output_delay_ms?`: integer - delay in milliseconds (0-5000), only set if `command` is `set_output_delay`
 
+The server MUST NOT send a player command to a client before that client has sent a [`client/state`](../../messaging.md#client--server-clientstate) containing the `player` object, which is where the client advertises `supported_commands`.
```
kahrendt: This should mirror the language used in https://github.com/Sendspin/spec/blob/5bea14e656bb0b5380ff5acb9e66412db5772d48/messaging.md?plain=1#L262 or maybe not even list it here since it is already covered in the messaging docs

roles/controller/v1.md
maximmaxim345: Good catch, moved it into the lead-in so all the steps inherit it.

roles/controller/v1.md
maximmaxim345: Dropped it instead of rewriting. It followed from the two rules above it.

roles/player/v1.md
maximmaxim345: Not covered actually, messaging.md only gates binary data. But I'll open a separate PR to fix that.

maximmaxim345: #204

kahrendt (approved):
Thanks!

PR #179, opened 2026-08-28
Restructure pair-method descriptors

arturpragacz:
Restructure pair-method descriptors.

- Key pair-method descriptors by `method`. A dictionary fits much better here than a list.
- Expose them in management. The issue we have currently is that if a server wants to show more information about a given pairing method it cannot do that until that method is enabled, as those are the only ones that `client/hello` message lists.


maximmaxim345 (approved):
Looks good, thanks @arturpragacz !

PR #183, opened 2026-08-28
Remove the management namespace

### LINKED ISSUE #182: Remove the management namespace from the spec
## Background

We evaluated the spec for legal compliance on the manufacturer side. Manufacturers will not accept the amount of control the current spec gives servers over clients: the `management/*` namespace lets a paired server remotely read and modify a client's stored pairing records and pairing configuration.

## Decision

Simplify the spec by removing the entire `management/*` namespace.

- Clients store pairing records with a minimum capacity of 5; more is allowed.
- When a pairing completes at capacity, the client evicts an existing record so pairing never fails for lack of storage. Which record is evicted is implementation-defined (for example, least recently used), except a record backing a currently-open connection is never evicted.
- An evicted server recovers through the existing Sentinel fallback: on its next connection it learns its credential is gone and can offer its operator re-pairing.

## Consequences

- All `management/*` messages and `management/result` are removed, along with storage accounting, `record_mode`, and the `used` flag.
- Shared-PSK records are removed; every pairing record is bound to a `server_id` (the stored-pubkey model is the only model).
- Remote rotation of pairing secrets, remote enable/disable of pairing methods, the runtime `unpaired_access` toggle, and remote pairing-window opening become local, manufacturer-defined actions.
- The `'management'` activity disappears from `server/activate`, the connection priority ranking, and the `client/goodbye` reasons.
- `server/unpair` is unchanged, minus its shared-record carve-out.

## Follow-up (separate work)

Remove management support from the implementations (aiosendspin, sendspin-cpp) and the conformance suite once the spec change lands.

balloobbot:
Fixes #182.

Removes the entire `management/*` namespace and replaces it with a fixed pairing-record model.

## Changes

- `management.md` deleted: all `management/*` commands, `management/result`, storage accounting, `record_mode`, and the `used` flag are gone.
- `pairing.md`: new **Pairing Records** section. A record is the long-term PSK bound to a `server_id`; re-pairing replaces a server's own record. Clients MUST store at least 5 records. A pairing that completes at capacity MUST succeed by evicting an existing record; the choice is implementation-defined (for example, least recently used), except the record backing a currently-open connection MUST NOT be evicted. Evicted servers recover through the existing Sentinel fallback. Method enable/disable, secret rotation, the `unpaired_access` toggle, and pairing-window opening are local manufacturer-defined actions now.
- `messaging.md`: `activities` is now `('playback' | 'pairing')[]`; the long-term-PSK allowed sets are `[]`, `['playback']`, `['pairing']`. `server/unpair` loses its shared-record carve-out. The `management/` prefix and the management cases in the `unauthorized` and `concurrent_attempt` goodbye reasons are removed.
- `connection.md`: connection priority is now `'playback'` > `'pairing'`. The shared-PSK storage model is removed; every record stores `server_id` and gets the post-match check. `psk_id` uniqueness is enforced at generation time. Eviction is added to the Sentinel-fallback causes.
- `README.md` regenerated; `build-readme.py --check` passes.

## Follow-up

The implementations (aiosendspin, sendspin-cpp) and the conformance suite still carry management support and need matching removals once this lands.


pairing.md
```diff
@@ -10,14 +10,24 @@ This specification defines three pairing methods. Servers must implement all thr
...
+
+### Pairing Records
 
-Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
+Each successful pairing produces a pairing record: the new [long-term PSK](README.md#definitions) persisted together with the server's `server_id`. A repeat pairing with a server that already holds a record replaces that record.
+
+A client MUST be able to store at least 5 pairing records; more is allowed. When a pairing completes at capacity, the client MUST evict an existing record so that the new record persists - a pairing never fails for lack of record storage. Which record is evicted is implementation-defined (for example, the least recently used), except that the client MUST NOT evict the record backing a currently-open connection.
```
maximmaxim345: @kahrendt What do you think about the 5 pairing records limit for ESPHome?
I think one of the reasons for the shared pairing records approach was to minimize storage usage on constraint devices.
This will require clients to reserve at least 320 bytes for Sendspin.

connection.md
```diff
@@ -118,25 +115,18 @@ Sentinel psk_id = 0x185b15f6d2da4909bd1dc156a4ab206103abef0153bcd52d926170b95cf7
...
+The server verifies the second handshake message against the PSK its first message referenced. If that fails and the referenced PSK was not the Sentinel, it verifies the same message against the Sentinel PSK before treating the handshake as failed. A second message that validates under the Sentinel is an authenticated **credential-mismatch signal**: the handshake authenticates the client's static key, so the signal proves its holder could not use the referenced PSK. The signal alone MUST NOT cause either side to remove or replace a record; records change only as described in [Pairing Records](pairing.md#pairing-records).
```
maximmaxim345: `server/unpair` also change the records (that messages behavior is not mentioned in `pairing.md#pairing-records`)
Factory reset too.
And manufacturer also cant touch the records with this current wording.

messaging.md
```diff
@@ -414,15 +413,14 @@ Sent by the client before gracefully closing the connection. This allows the cli
...
   - `shutdown` - client is shutting down. When the device is powering off or otherwise not coming back and no more specific reason applies, clients SHOULD send this reason. Server should not auto-reconnect
   - `restart` - client is restarting and will reconnect. Server should auto-reconnect
   - `user_request` - user explicitly requested to disconnect from this server. Server should not auto-reconnect
-  - `unauthorized` - the client is no longer authorized for the connection: either the server declared an activity set the client is not authorized for (e.g., `'management'` on an [unpaired](README.md#definitions) session), or the client removed its own pairing record (see [`management/remove-record`](management.md#server--client-managementremove-record)) and can no longer authenticate. Server should not auto-reconnect with the same activity set
+  - `unauthorized` - the server declared an activity set or `active_roles` the client is not authorized for (see [`server/activate`](#server--client-serveractivate)). Server should not auto-reconnect with the same activity set
```
maximmaxim345, line 421: This dropped the "client removed its own pairing record" case, that is still possible though, just not via a management message.

pairing.md
```diff
@@ -10,14 +10,24 @@ This specification defines three pairing methods. Servers must implement all thr
...
 3. **Static Pairing Code** - pairing with a fixed [Pairing Code](README.md#definitions). Appropriate for devices with no out-channel; vulnerable to MITM if the pairing code is disclosed. See [Static Pairing Code Flow](#static-pairing-code-flow).
 
-- **Unpaired.** Sentinel PSK; the channel is unauthenticated until the CPace round completes. The round establishes trust from scratch and produces a new [long-term PSK](README.md#definitions).
+A code-based pairing starts from one of two states:
+
+- **Unpaired.** The handshake used the Sentinel PSK; the channel is unauthenticated until the [PAKE](#pake) round completes. The round establishes trust from scratch and produces a new [long-term PSK](README.md#definitions).
 - **Already paired.** The server moves the established connection into pairing (see [Entering and leaving pairing](#entering-and-leaving-pairing)) and runs the round over the existing long-term PSK.
```
maximmaxim345: Question: Should we still keep this already paired case?

pairing.md
```diff
@@ -10,14 +10,24 @@ This specification defines three pairing methods. Servers must implement all thr
...
+
+Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code`, which is intended for devices without one. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
+
+### Pairing Records
 
-Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
+Each successful pairing produces a pairing record: the new [long-term PSK](README.md#definitions) persisted together with the server's `server_id`. A repeat pairing with a server that already holds a record replaces that record.
```
```suggestion
Each successful pairing produces a pairing record: the new [long-term PSK](README.md#definitions) persisted together with the server's `server_id`. The client MUST persist the new record, replacing any record it already holds for the server.
```

pairing.md
maximmaxim345: Provisional connections can also be fully authenticated, was it ment to include those in the eviction too? If yes this should close the edge case:
```suggestion
A client MUST be able to store at least 5 pairing records; more is allowed. When a pairing completes at capacity, the client MUST evict an existing record so that the new record persists - a pairing never fails for lack of record storage. Which record is evicted is implementation-defined (for example, the least recently used), except that the client MUST NOT evict a record backing a currently-open connection, provisional or admitted; the client MUST cap its concurrently open paired connections below its record capacity (see [Multiple servers](connection.md#multiple-servers-server-initiated)) so an evictable record always exists.
```

pairing.md
```diff
@@ -35,7 +45,7 @@ A server MAY send such a cancelling `server/activate` at any point during a pair
 
 ### Unpaired Access
 
-A client MAY admit a server with no pairing record to activate roles or declare the `'playback'` activity. The session is [unpaired](README.md#definitions), so [management](management.md#management) operations remain unavailable. Whether a client admits unpaired access is governed by its `unpaired_access` setting: the default is the manufacturer's choice, the toggle is exposed at runtime via [`management/set-pairing-config`](management.md#server--client-managementset-pairing-config), and the current value is advertised in [`client/hello`](messaging.md#client--server-clienthello) as `unpaired_access.enabled`.
+A client MAY admit a server with no pairing record to activate roles or declare the `'playback'` activity. The session is [unpaired](README.md#definitions). Whether a client admits unpaired access is governed by its `unpaired_access` setting: the default is the manufacturer's choice, changing it is a local client action (manufacturer-defined), and the current value is advertised in [`client/hello`](messaging.md#client--server-clienthello) as `unpaired_access.enabled`.
```
maximmaxim345: Hmm, I think if a client disables this though a manufacturer defined local action during the playback activity this will cause undefined behavior.

pairing.md
```diff
@@ -47,7 +57,7 @@ While a client is unapproved, the server SHOULD identify and present it to the o
...
+**Lifecycle.** The client's pairing PSK MUST be drawn from a [CSPRNG](README.md#definitions) per device and MUST NOT be a fixed default shared across devices, whether provisioned at manufacture or generated by the client. It persists across reboots and is per-client and long-lived: a successful pairing does not consume it (pairing produces a separate [long-term PSK](README.md#definitions)), so it can pair the client with any number of servers.
```
maximmaxim345, line 57: Dropping the rotation prose also dropped "MUST NOT rotate it on its own", so nothing now forbids a client rotating this spontaneously. Just double checking, is that intentional to keep it completely up to the manufacturer?

pairing.md
```diff
@@ -243,7 +253,7 @@ Tokens are drawn only from the QR code alphanumeric set (`0–9`, `A–Z`, `:`),
 Decoding reverses the transform and MUST be lenient with operator-supplied input:
 
 1. Trim surrounding whitespace and upper-case it.
-2. If present, strip a leading `SP:`. The first character is the `version`; reject an unrecognized version. An interface that accepts several versions dispatches on this character; one expecting a specific version rejects the others.
+2. If present, strip a leading `SP:`. The first character is the `version`; reject an unrecognized version.
```
maximmaxim345: This dropped the only place saying that an interface expecting a specific version rejects the others.
Now scanning the printed version 0 token into a `qr_code` prompt is undefined.

pairing.md
```diff
@@ -259,7 +269,7 @@ An attempt is **gesture-gated** - the client withholds [`client/pair-init`](#cli
 
 Pairing Window mechanics:
 
-- **Opening the window.** An operator gesture on the client - a physical button press, a reset-pinhole press, a button combo, a specific power-cycle pattern, a shake or motion gesture, or any equivalent implementation-defined action - or a paired server via [`management/open-pairing-window`](management.md#server--client-managementopen-pairing-window). Gestures SHOULD be deliberate and hard to induce remotely.
+- **Opening the window.** An operator gesture on the client - a physical button press, a reset-pinhole press, a button combo, a specific power-cycle pattern, a shake or motion gesture, or any equivalent implementation-defined action. Gestures SHOULD be deliberate and hard to induce remotely.
```
maximmaxim345, line 267: I think we should now require a client implementing `dynamic_pairing_code` to have one. Otherwise it can get locked out permanently (other than a factory reset) since we don't even have `management/open-pairing-window` now.

messaging.md
```diff
@@ -192,7 +192,7 @@ Players that can output audio should have the role `player`.
...
 - `artwork@v1_support?`: object - required if `artwork@v1` is listed, absent otherwise ([see artwork@v1 support object details](roles/artwork/v1.md#client--server-clienthello-artworkv1-support-object))
 - `visualizer@v1_support?`: object - required if `visualizer@v1` is listed, absent otherwise ([see visualizer@v1 support object details](roles/visualizer/v1.md#client--server-clienthello-visualizerv1-support-object))
-- `supported_pair_methods`: object - pairing methods this client currently offers, keyed by method identifier, each value a [pair-method descriptor](pairing.md#client--server-clienthello-pair-method-descriptor). An implemented method that is [disabled](management.md#server--client-managementset-pairing-config) is omitted.
+- `supported_pair_methods`: object - pairing methods this client currently offers, keyed by method identifier, each value a [pair-method descriptor](pairing.md#client--server-clienthello-pair-method-descriptor). Every client implements at least the Pairing PSK method (see [Pairing](pairing.md#pairing)).
```
maximmaxim345: I think this is slightly underdefined now, with enable/disable gone it has to go one of two ways.
If manufacturers can still change the offered methods at runtime, `pairing_psk` needs to be pinned as always present here (dropping locked_down relies on that).
If not, "currently offers" should just say implements.

maximmaxim345 (commented):
Previously runtime changes to the pairing config went through management, which needed the admitted connection, so the server that changed the state was also the one that knew the fresh state, and every change had a defined result.

Now local actions can change it at any time, but `client/hello` is only sent on a new connection and the client has no way to announce a change: a server that got `pairing_required` never learns that unpaired access was enabled later.
I think we should solve this at least somehow. For connected servers the simplest solution is to require the client to send a `client/goodbye` with `restart` after such a change, but I'm not sure if that is the best solution. If a server was previously disconnected through `pairing_required`, it has no connection left to receive a goodbye...

Also, @arturpragacz should also carefully review this PR before this is merged.

arturpragacz (commented):
This should be addressed first: https://github.com/Sendspin/spec/issues/182#issuecomment-5476990375

pairing.md
arturpragacz: I'd say yes, arguably even more so now, as it becomes the only way to rotate the credential by the server.

connection.md
balloob: It's the goal for manufacturers to not touch the records, except for `server/unpair` to remove it, which requires an authenticated connection.

messaging.md
balloob: When a server sents `server/unpair`, the client sents a `client/goodbye` with reason `unpaired`.

pairing.md
balloob: I think that we can remove it. To rotate a credential a server can unpair the existing connection and then initiate pairing again. 

What would be the reason a server would want to rotate their keys? This is a feature that we can consider adding back later if there is demand.

pairing.md
balloob: Yep. We do not want to mandate how the client handles this.

pairing.md
balloob: You can always turn a device off and on again.

connection.md
balloobbot: Fixed in c74d6d0

pairing.md
balloobbot: Removal done in #192

pairing.md
balloobbot: Fixed in c74d6d0

pairing.md
balloobbot: Fixed in c74d6d0

pairing.md
balloobbot: Fixed in c74d6d0

pairing.md
balloob: Should we add that client sends `client/goodbye` with `pairing_required` ?

pairing.md
balloob: Correct. For a client that is similar to dropping a pairing record, as the server won't be able to pair. I don't think it's worth mentioning. Just like a client can't start rewriting server_id, or mutate other data.

messaging.md
maximmaxim345: Makes sense with https://github.com/Sendspin/spec/pull/183#discussion_r3898330499 as context. Resolving.

pairing.md
maximmaxim345: That should solve it.
The `pairing_required` description in `client/goodbye` needs a small update then:
> the client refused, or no longer admits, an [unpaired access](pairing.md#unpaired-access) connection because it does not have unpaired access enabled. Server should not auto-reconnect without pairing first

pairing.md
maximmaxim345: Turning it off and on doesn't reset anything, the failure counter is required by the spec to be persisted across reboots.
A power cycle only opens a window if the manufacturer implemented a power-cycle pattern as their gesture. A simple off and on would technically work since "deliberate and hard to induce remotely" is only a SHOULD, but still goes against the spec's recommendation.

pairing.md
balloobbot: Fixed in 2a2b9c2

pairing.md
balloobbot: Replacing escalation with time-based backoff (and removing the persistent counter) in #194

messaging.md
balloob: - Drift is already handled at the point of use (`pair/abort` with `method_not_supported` when a server picks something no longer offered).

- `pairing_psk` is already pinned by the existing sentence "every client implements at least the Pairing PSK method."

maximmaxim345: If the offered set can drift (point 1), then a fixed property like implements can't pin what's offered. One word fixes it: "offers" instead of "implements" in that sentence.

balloob:
```suggestion
- `supported_pair_methods`: object - pairing methods this client currently offers, keyed by method identifier, each value a [pair-method descriptor](pairing.md#client--server-clienthello-pair-method-descriptor). Every client offers at least the Pairing PSK method (see [Pairing](pairing.md#pairing)).
```

pairing.md
kahrendt: Storing at least 5 records will not be an issue on an esp32, so this is fine in my view. An esp8266 would have a harder time (but probably still could), but we've never targeted them in any way, so it shouldn't guide us here.

marcelveldt (approved):

arturpragacz:
Just for the future record: I personally am not in favour of this change.

PR #187, opened 2026-08-28
Drop unpaired-access shipping default for microphone clients

balloobbot:
Removes the shipping-default recommendation added in #170:

> Clients with a privacy-sensitive input such as a microphone SHOULD ship with unpaired access disabled: the captured audio would otherwise be accessible to anyone on the network.

A device can ship with Sendspin off and enable it later through its own onboarding; that onboarding then owns the unpaired-access decision. Because this is client responsibility, the spec should not prescribe a shipping default.

The server-side protection is unchanged: activating `source@v1` on an unpaired connection still requires explicit operator approval, never implied approval.


maximmaxim345 (commented):
Still not sure if removing it completely is a good idea.

How about re-adding a note that anybody can access the source (with the example microphone) on the network? It's still the implementors decision what to do with this information then.

For reference: `pairing.md:44` for example mentions MITM attacks just to inform the implementor too.

balloob:
Yeah, in hindsight, let's keep it like this.

PR #188, opened 2026-08-28
Split large artwork images across multiple messages

### LINKED ISSUE #134: Artwork transfers can starve audio: head-of-line blocking on the shared connection

kahrendt:
An artwork image larger than one Noise frame (65518 bytes of payload) currently relies on transport fragmentation, which admits no interleaving: the sender must finish the fragmented message before any other frame, so the whole image is an indivisible burst that audio chunks cannot preempt. A player holding only `min_buffer_ms` of audio can underrun behind it; on a live stream that truncates audio. With BMP gone (#168) the worst case shrank, but photographic JPEG still exceeds one frame beyond roughly 500x500 (the spec's own `stream/start` example, 800x800 JPEG, is 3-5 frames), and PNG exceeds it at almost any artwork size.

## Changes:

- An image is transferred as a fixed 14-byte announce message carrying the timestamp and the image's `total_size` (uint32), followed by parts carrying only image data, all on the channel's message type; a flags byte after the type byte distinguishes the message kinds, following the shape #172 gave fragmentation
- The announce carries no image data, so a client can allocate its image buffer before any image bytes arrive and write each part directly into it; the transfer completes when the received data reaches `total_size`
- Every artwork message is capped at 65518 bytes after the type byte so it never needs transport fragmentation; any other messages MAY be sent between the messages of a transfer, and transfers on different channels are independent
- The pending image is now the most recently announced image, from its announce until it becomes current: an announce discards any held pending image, partly received or complete, and the pending image becomes current once its transfer completes and its timestamp is reached
- A rejected announce (no active stream, or client not available) still starts its transfer, so its parts are rejected rather than treated as malformed
- The clear message becomes an announce with `total_size` 0, which completes immediately with no parts
- A two-byte cancel message (flags bit 1) discards the channel's pending image, partly received or complete, leaving the current image showing; it replaces resend-the-current-image as the cancel recipe in the scheduled-artwork rules, which under chunking would cost a full image retransfer
- Malformed sequences (a message under 2 bytes or over the frame cap, an announce whose length is not 14 bytes, a cancel longer than 2 bytes, a part with no transfer in flight, data extending past `total_size`, a nonzero reserved flag bit, both flag bits set) close the connection, matching fragmentation

The pending-image redefinition keeps the newest-wins rule from #135 applying at the announce, as it did at message arrival before chunking, and means a client needs at most two image buffers per channel, receiving the transfer directly into the pending one.

Every image costs one extra small message, including images that fit in a single frame; the uniform format was judged worth ~31 bytes of ciphertext per image to give clients flexibility to transfer directly into a proper sized buffer.

There is no explicit last-part flag; `total_size` already determines completion, and a redundant end marker would add mismatch states to define.

Every artwork message changes shape: the flags byte shifts the timestamp, the header and image data travel in separate messages, and the empty clear message becomes a bare announce with `total_size` 0.

Closes #134

## Breaking changes:

The wire format for artwork is completely changed. All servers and clients will have to re-implement to handle:

- the announce binary message
- the artwork specific split chunks versus using the protocol's generalized noise split
- handling the pending slot; i.e., servers should now send a cancel message instead of re-sending the current image


roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
...
+- `flags` (byte 1): uint8 - bit 0 is set on an announce and bit 1 on a cancel message (defined below); a part sets neither. Bits 2-7 are reserved and MUST be zero
+- `timestamp` (announce, bytes 2-9): big-endian int64 - server clock time in microseconds when the image should be displayed by the device
+- `total_size` (announce, bytes 10-13): big-endian uint32 - size in bytes of the encoded image
+- `data` (part, rest of bytes): the next bytes of the encoded image
+
+The concatenated `data` of all parts is the encoded image; the transfer is complete when the received data reaches `total_size`. An announce with `total_size` `0` completes immediately, with no parts. An artwork message, after the type byte, MUST NOT exceed 65518 bytes, so that it fits in a single Noise transport message without [fragmentation](../../messaging.md#fragmentation).
+
+Per channel, at most one image transfer is in flight at a time: a transfer is in flight from its announce until it completes or its partly received image is discarded. Transfers on different channels are independent, and any other messages MAY be sent between the messages of a transfer.
```
maximmaxim345: Should we instead limit it to at most image transfer for the entire `artwork@v1` role instead?
Otherwise we can still have up to 4 simultaneous data streams (plus anything else by other Sendspin roles).

roles/artwork/v1.md
maximmaxim345: Can you adjust the "immediately" to "promptly"? Since we now serialize announces after 2d033b166ed3549208b4c72a7eb1ae1414f0dd86.

roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
 
 ### Server → Client: Artwork (Binary)
 
-Binary messages SHOULD be rejected if there is no active stream or the client is not [`available`](../../messaging.md#client--server-clientstate).
+Binary messages SHOULD be rejected if there is no active stream or the client is not [`available`](../../messaging.md#client--server-clientstate). A rejected announce still starts its transfer, whose parts are likewise rejected.
```
maximmaxim345: This only covers rejection at the announce, but a transfer can also start while available and then continue after the client goes `available: false`.
If a client tears the transfer down there, the remaining parts hit the "no transfer in flight" rule and close the connection.

roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
...
+- Announce: `[type][flags][timestamp][total_size]`
+- Part: `[type][flags][data]`
+- Cancel: `[type][flags]`
 
-**Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty binary message (only the message type byte and timestamp, with no image data) for that channel. An empty message follows the same rules as any other image: a future timestamp schedules the clear.
+The fields are:
+
+- `type` (byte 0): uint8, `8`-`11` - corresponds to artwork channel 0-3 respectively
+- `flags` (byte 1): uint8 - bit 0 is set on an announce and bit 1 on a cancel message (defined below); a part sets neither. Bits 2-7 are reserved and MUST be zero
```
maximmaxim345: Very tiny nit: I think we should swap them so a future role that needs parts but no announce can be more consistent with this role and looks prettier.

roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
...
+- `timestamp` (announce, bytes 2-9): big-endian int64 - server clock time in microseconds when the image should be displayed by the device
+- `total_size` (announce, bytes 10-13): big-endian uint32 - size in bytes of the encoded image
+- `data` (part, rest of bytes): the next bytes of the encoded image
+
+The concatenated `data` of all parts is the encoded image; the transfer is complete when the received data reaches `total_size`. An announce with `total_size` `0` completes immediately, with no parts. An artwork message, after the type byte, MUST NOT exceed 65518 bytes, so that it fits in a single Noise transport message without [fragmentation](../../messaging.md#fragmentation).
```
maximmaxim345: Nit: The cap is measured after the type byte but the malformed rule reads whole-message, like the three other length rules next to it, so a legal 65519 byte message looks malformed.
```suggestion
The concatenated `data` of all parts is the encoded image; the transfer is complete when the received data reaches `total_size`. An announce with `total_size` `0` completes immediately, with no parts. An artwork message MUST NOT exceed 65519 bytes, so that it fits in a single Noise transport message without [fragmentation](../../messaging.md#fragmentation).
```

roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
...
+
+The timestamp indicates when this artwork should be displayed. Per channel, clients keep the **current image**, which is always what the channel shows, plus at most one **pending image**: the most recently announced image, from its announce until it becomes current. An announce discards any held pending image. The pending image becomes current once its transfer is complete and its timestamp, translated to the local clock via the [time filter](../../messaging.md#clock-synchronization) (current best estimate, no waiting for convergence), has been reached; artwork is never dropped for lateness. Clients MAY ease into a complete pending image around its timestamp (e.g. a cross-fade) or show it early (e.g. a coming-up display). On [`stream/end`](../../messaging.md#server--client-streamend), clearing buffers includes discarding pending images. A [`stream/start`](../../messaging.md#server--client-streamstart) that changes a channel's configuration likewise discards that channel's pending image, and the server re-sends the image if it still applies.
```
maximmaxim345: Should "any held" be "that channel's"?
Now that transfers are role-global this reads as discarding the pending image on every channel, so taken literally a 4 channel track change would leave only the last channel showing.

maximmaxim345: Line 79 says a server cancels a transfer first to replace an image, but nothing says what happens to one when a `stream/end` or `stream/start` arrives.

I think the server should be responsible to cancel the transfer before any `stream/end` or `stream/start` (or only the ones changing artwork config?), so the client has one teardown path instead of deriving it from three messages.

roles/artwork/v1.md
```diff
@@ -56,24 +56,38 @@ Each channel's configuration MUST match the client's current capability for that
...
+
+**Clearing artwork:** To clear the currently displayed artwork on a specific channel, the server sends an empty image for that channel: an announce with `total_size` `0`. An empty image follows the same rules as any other image: a future timestamp schedules the clear.
+
+**Cancel message:** A message consisting of only the `type` and `flags` bytes, with bit 1 set. It discards the channel's pending image, taking effect immediately; the current image is unaffected.
+
+**Malformed sequences** are protocol errors; the client MUST close the connection. They are: a message shorter than 2 bytes or exceeding the size cap above, an announce whose length is not 14 bytes, an announce received while a transfer is in flight, a cancel message longer than 2 bytes, a part received with no transfer in flight or on a channel other than the in-flight transfer's, a part whose `data` would extend past `total_size`, a nonzero reserved flag bit, and a message with bit 0 and bit 1 both set.
 
 #### Server rules for scheduled artwork
 
 Servers SHOULD NOT send a scheduled image more than 20 seconds before its timestamp.
```
maximmaxim345: Now that an image is several messages, it is not clear which one the 20 seconds counts from.
I think it should be the announce, since that is when the client starts holding the buffer.

roles/artwork/v1.md
maximmaxim345, line 21: Just noticed (was pre-existing): When the server sets a channel to `none` itself, the pending image is discarded but nothing clears the current one, so the channel keeps showing a stale image.
Should the server send an empty image for it first?

roles/artwork/v1.md
kahrendt: https://github.com/Sendspin/spec/pull/188/commits/d435873bec66fd38e416ea6429e0653f3b7dc7a5

roles/artwork/v1.md
kahrendt: https://github.com/Sendspin/spec/pull/188/commits/dc289fea19766ecb43de39c3dcac532cbad65a99

roles/artwork/v1.md
kahrendt: https://github.com/Sendspin/spec/pull/188/commits/3b9c9b3c4f6c6cdb1713d7a65a720480845da280

roles/artwork/v1.md
kahrendt: Good catch! https://github.com/Sendspin/spec/pull/188/commits/5999727d871e7f399a9bbbd0526e9ca42dfc5158

roles/artwork/v1.md
kahrendt: https://github.com/Sendspin/spec/pull/188/commits/599b82ba2f4874a93923a30f82aa357faba30b7f

roles/artwork/v1.md
kahrendt: I specified the server should send an empty image first: https://github.com/Sendspin/spec/pull/188/commits/fc77979156f961f5bf1a36cb17e1b9c6b0ff1dfe

roles/artwork/v1.md
kahrendt: I tried to spell it out more carefully here: https://github.com/Sendspin/spec/pull/188/changes/10329755736d80f79a468dd0f5cada2ff228c6f1
Clients still need to monitor and track stuff as it comes in because they otherwise cannot determine when a message is finished.

roles/artwork/v1.md
kahrendt: https://github.com/Sendspin/spec/pull/188/changes/4c148e1fa5a95f5f49758413e71c24272adbc067

I went with the server is responsible for cancelling it before sending `stream/end` or `stream/start` with a configuration schange.

roles/artwork/v1.md
```diff
@@ -52,28 +52,42 @@ The `channels` array covers every channel index the client declared in [`artwork
...
+- `total_size` (announce, bytes 10-13): big-endian uint32 - size in bytes of the encoded image
+- `data` (part, rest of bytes): the next bytes of the encoded image
+
+The concatenated `data` of all parts is the encoded image; the transfer is complete when the received data reaches `total_size`. An announce with `total_size` `0` completes immediately, with no parts. An artwork message MUST NOT exceed 65519 bytes, so that it fits in a single Noise transport message without [fragmentation](../../messaging.md#fragmentation).
+
+At most one image transfer is in flight at a time across all of the role's channels: a transfer is in flight from its announce until it completes or its partly received image is discarded. The server MUST NOT announce an image, on any channel, while a transfer is in flight; to replace an image still in flight, it cancels that transfer first. The server MUST cancel an in-flight transfer before sending a [`stream/end`](../../messaging.md#server--client-streamend) or a [`stream/start`](../../messaging.md#server--client-streamstart) that changes that transfer's channel configuration. Any other messages MAY be sent between the messages of a transfer.
```
maximmaxim345:
```suggestion
At most one image transfer is in flight at a time across all of the role's channels: a transfer is in flight from its announce until it completes or its partly received image is discarded. The server MUST NOT announce an image, on any channel, while a transfer is in flight; to replace an image still in flight, it cancels that transfer first. The server MUST cancel an in-flight transfer before sending an artwork [`stream/end`](../../messaging.md#server--client-streamend) or a [`stream/start`](../../messaging.md#server--client-streamstart) that changes that transfer's channel configuration. Any other messages MAY be sent between the messages of a transfer.
```

roles/artwork/v1.md
```diff
@@ -52,28 +52,42 @@ The `channels` array covers every channel index the client declared in [`artwork
...
+- `total_size` (announce, bytes 10-13): big-endian uint32 - size in bytes of the encoded image
+- `data` (part, rest of bytes): the next bytes of the encoded image
+
+The concatenated `data` of all parts is the encoded image; the transfer is complete when the received data reaches `total_size`. An announce with `total_size` `0` completes immediately, with no parts. An artwork message MUST NOT exceed 65519 bytes, so that it fits in a single Noise transport message without [fragmentation](../../messaging.md#fragmentation).
+
+At most one image transfer is in flight at a time across all of the role's channels: a transfer is in flight from its announce until it completes or its partly received image is discarded. The server MUST NOT announce an image, on any channel, while a transfer is in flight; to replace an image still in flight, it cancels that transfer first. The server MUST cancel an in-flight transfer before sending an artwork [`stream/end`](../../messaging.md#server--client-streamend) or a [`stream/start`](../../messaging.md#server--client-streamstart) that changes that transfer's channel configuration. Any other messages MAY be sent between the messages of a transfer.
```
maximmaxim345, line 79: Ok, last nit: I think we are missing a paragraph explaining why a server should not send all the parts back to back (so essentially continuing like before this PR).
Artwork is not time critical, so yielding between parts keeps a large transfer from blocking the other roles on the connection, like `player@v1` audio chunks.

maximmaxim345 (approved):
Looks good now (I have only one last nit, but that's more to give server implementers a hint so they avoid the pitfall this tries to solve).
Thanks @kahrendt!

roles/artwork/v1.md
kahrendt: Added a paragraph: https://github.com/Sendspin/spec/pull/188/commits/b20f9ed21a43c421ac389ed9800d0c0df439865d

kahrendt:
@maximmaxim345 Thanks for the thorough feedback on this one, it improved the PR greatly!

maximmaxim345 (approved):

PR #189, opened 2026-08-28
Allow at most one pairing-code method per client

### LINKED ISSUE #184: Decide: keep or trim the pairing-method steering
pairing.md currently steers method choice:

> Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code`, which is intended for devices without one. Clients whose display can render a QR code SHOULD also offer the `qr_code` emission format.

Decide whether to keep this steering or trim it: which pairing methods to implement is arguably the client's decision, and the spec could limit itself to describing the methods and their trade-offs (the MITM note on static codes already does that).

Related: #186 (at most one pairing-code method per client) - if that lands, this paragraph changes anyway.

### LINKED ISSUE #186: Decide: allow at most one pairing-code method per client
Today a client must implement Pairing PSK and may additionally implement either or both pairing-code methods (`dynamic_pairing_code`, `static_pairing_code`).

Decide whether to enforce that a client offers at most one pairing-code method: static or dynamic, not both.

For: a device with a working out-channel has no need for a static code (weaker: MITM if the code leaks), servers get a simpler method choice, and clients drop one flow plus its gating policy.

Against: no fallback when the out-channel is unusable (broken display, muted speaker), and a printed static code is the only method that works with the out-channel occupied.

No wire change either way - `supported_pair_methods` already carries whatever the client offers; this would only constrain what may appear there.

balloobbot:
Fixes #184. Fixes #186.

Two changes to pairing.md's Methods section:

- Clients may implement at most one pairing-code method: static or dynamic, not both (Pairing PSK stays mandatory). No wire change - `supported_pair_methods` already carries whatever the client offers; this constrains what may appear there.
- The steering paragraph (SHOULD implement `dynamic_pairing_code` given an out-channel, SHOULD offer `qr_code` given a display) is changed to a regular non binding "should". Method choice is the client's decision; the method list already describes the trade-offs (out-channel requirement, static-code MITM caveat).

Note: overlaps textually with #183, which edits the same paragraph - whichever lands second needs a trivial rebase.

### DIFF
--- a/pairing.md
+++ b/pairing.md
@@ -2,7 +2,7 @@
 
 Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](connection.md#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](connection.md#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](connection.md#re-handshake) to the newly delivered `long_term_psk`, promoting the channel to a paired session without closing the WebSocket.
 
-This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement either or both pairing-code methods.
+This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement at most one pairing-code method: static or dynamic, not both.
 
 ### Methods
 
@@ -15,8 +15,6 @@ A code-based pairing runs over a Sentinel-keyed connection: the channel is unaut
 The client reveals the new long-term PSK only after `server_kc` verifies, and only as `wrapped_psk` [sealed under the CPace output](#wrapping): a peer that cannot complete the PAKE - wrong pairing code, or a man in the middle relaying between two handshakes, whose differing `h` gives each leg a different `sid` - neither triggers the reveal nor can unwrap it.
 
 Static pairing methods (Pairing PSK, Static Pairing Code) do not take over the device's out-channel. Dynamic pairing (Dynamic Pairing Code) takes over the out-channel - typically the audio output or display - to emit the per-session pairing code, so it cannot run while audio is playing on the same device; the operator must stop playback before initiating pairing (see [Multiple servers](connection.md#multiple-servers-server-initiated)).
-
-Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code`, which is intended for devices without one. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
 
 ### Pairing Records
 
--- a/pairing.md
+++ b/pairing.md
@@ -2,7 +2,7 @@
 
 Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](connection.md#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](connection.md#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](connection.md#re-handshake) to the newly delivered `long_term_psk`, promoting the channel to a paired session without closing the WebSocket.
 
-This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement at most one pairing-code method: static or dynamic, not both.
+This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement at most one pairing-code method: Static Pairing Code or Dynamic Pairing Code, not both.
 
 ### Methods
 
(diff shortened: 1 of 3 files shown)


pairing.md
```diff
@@ -2,7 +2,7 @@
 
 Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](connection.md#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](connection.md#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](connection.md#re-handshake) to the newly delivered `long_term_psk`, promoting the channel to a paired session without closing the WebSocket.
 
-This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement either or both pairing-code methods.
+This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement at most one pairing-code method: static or dynamic, not both.
```
maximmaxim345: I think this also belongs in the advertisement (like: "a client MUST NOT list both in `supported_pair_methods`")
And same server side rule so a client announcing both doesn't cause undefined behavior.

maximmaxim345: Call this "Static Pairing Code or Dynamic Pairing Code" for better ctrl-f searchability.

pairing.md
```diff
@@ -17,8 +17,6 @@ The client reveals the new long-term PSK only after `server_kc` verifies, and on
 
 Static pairing methods (Pairing PSK, Static Pairing Code) do not take over the device's out-channel. Dynamic pairing (Dynamic Pairing Code) takes over the out-channel - typically the audio output or display - to emit the per-session pairing code, so it cannot run while audio is playing on the same device. A pairing attempt that arrives while another connection is playing is rejected (see [Multiple servers](connection.md#multiple-servers-server-initiated)); the operator must stop playback before initiating pairing.
 
-Clients with a usable out-channel (display, speaker, etc.) SHOULD implement `dynamic_pairing_code` and prefer it to `static_pairing_code` - but SHOULD implement `static_pairing_code` too, shipped [disabled](management.md#server--client-managementset-pairing-config) with no pairing code provisioned. Clients whose display can render a QR code SHOULD also offer the `qr_code` [emission format](#dynamic-pairing-code-flow).
```
maximmaxim345: I guess this will land in a future amendment/implementation guide?

pairing.md
balloob: Since PSK is always enabled. Maybe we can reduce pairing mode to just be a string?

pairing.md
balloob: Yes

pairing.md
maximmaxim345: I think right now PSK can be disabled through management.

And how will you then send the audio format for the digit audio pack?

pairing.md
arturpragacz: What's the reason for removing it from the spec though? I feel like this is a good advice to have here. And since it's only recommended, not required, if there's an important reason, then the implementation may deviate from it.

balloob: It's not recommended, it's a SHOULD

arturpragacz: SHOULD means the same as RECOMMENDED in RFC 2119 terms.

```
SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

pairing.md
balloob: All management commands are being removed in #183.

pairing.md
balloob: Ah yes, you're right. However, a `SHOULD` has more weight to it than just a recommendation. It will be flagged as a deviation that needs to be justified. But the core of this change is that the method to implement is a product decision, not a protocol matter. That's why it is better fit in the implementation guide.

pairing.md
balloob: Actually, no longer in favor of reducing pairing mode to a string. If we ever add NFC pairing, clients might support two again.

pairing.md
```diff
@@ -2,7 +2,7 @@
 
 Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](connection.md#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](connection.md#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](connection.md#re-handshake) to the newly delivered `long_term_psk`, promoting the channel to a paired session without closing the WebSocket.
 
-This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement either or both pairing-code methods.
+This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally implement at most one pairing-code method: Static Pairing Code or Dynamic Pairing Code, not both.
```
maximmaxim345: Should this cap implementing, or offering? Line 303 only caps what is listed, so the two do not exactly match. I wouldn't cap implementing since the SDK might implement both...
```suggestion
This specification defines three pairing methods. Servers must implement all three; clients must implement Pairing PSK and may additionally offer at most one pairing-code method: Static Pairing Code or Dynamic Pairing Code.
```

pairing.md
maximmaxim345: Can we at least leave this paragraph here until we have an implementation guide? We can drop the `SHOULD`s to `should` (and optionally prefix with `**Note:**`, or rewords to use non-RFC "recommended"), which addresses the "flagged as a deviation" concern IMO.

marcelveldt: yes this paragraph should not be removed - we still require them to implement pairing

maximmaxim345: Re-added it with 9dd6d11. But this time without the RFC type SHOULD (and regular should instead).

maximmaxim345 (approved):

PR #192, opened 2026-08-31
Remove the already-paired pairing entry

balloobbot:
Removes the already-paired entry into code-based pairing: a code-based pairing now runs over a Sentinel-keyed connection only, and the long-term PSK no longer admits the `'pairing'` activity.

A server has no need to run a pairing round over an existing credential: to establish a fresh one it unpairs (`server/unpair`) and pairs again. The Pairing PSK flow is unaffected - it still enters from any open connection by re-handshaking to the pairing PSK first, since that is an operator-token-driven fresh pairing.

Device-presence verification rode exclusively on this entry and is removed with it. If demand appears, it can come back as its own feature.

Split out of #183 (see [the pairing.md discussion](https://github.com/Sendspin/spec/pull/183#discussion_r3892781993)) so it can be discussed on its own. Overlaps textually with #183; whichever merges second needs a rebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,7 +509,7 @@ The activity sets the server may legitimately declare are constrained by which P
 
 | PSK matched | Allowed activity sets |
 |---|---|
-| [long-term PSK](#definitions) | `['pairing']` or any subset of `{'playback', 'management'}` |
+| [long-term PSK](#definitions) | any subset of `{'playback', 'management'}` |
 | [pairing PSK](#definitions) | `['pairing']` |
 | [Sentinel PSK](#pre-shared-key) | `[]`, `['pairing']`, `['playback']`¹ |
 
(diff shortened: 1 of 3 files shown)


maximmaxim345 (approved):
Was thinking a bit more of whether we should keep or drop key rotations.
I think use of key rotations is more theoretical than practical:
- If you know you have a compromised client, the best mitigation is not to rotate the key (since you can accidentally just rotate the key of the impostor) but unpairing it from the server, and probably also updating/factory resetting the client.
- What I can see as a proper process is to regularly rotate keys, that will solve part of my first point and will in addition be probably even more secure. But that comes at a severe disadvantage for low cost clients: each rotated key means rewriting persistent storage, causing increased wear.

PR #194, opened 2026-09-01
Rework pairing attempt gating

balloobbot:
Reworks how code-based pairing attempts are gated, from the [#183 lockout discussion](https://github.com/Sendspin/spec/pull/183#discussion_r3893019825).

Dynamic pairing code: the persisted failure counter and its escalation to gesture-gating are gone. A failed attempt is one where the client started emitting the code and the attempt ended without a successful `server_kc` verification. After 20 consecutive failed attempts the client MUST hold attempts back until a deliberate, manufacturer-defined operator action; which action is the manufacturer's choice, with guidance tracked in #205. The count is global across servers and addresses and resets on success or on that action. How a client paces attempts before the limit is its own choice, and it MAY be stricter, including a gesture from the first attempt. Since any deliberate action qualifies, no device is locked out, and since a client with no limit is not conformant, manufacturers have a spec section to point a test lab at (ETSI EN 303 645 provision 5.1-5, EN 18031-1 AUM-6).

Static pairing code: the window admits up to 5 attempts instead of one, all on the connection that carries the first, so a typo no longer costs a walk to the device. It closes on a completed pairing, the fifth failed attempt, drop of that connection, operator cancellation, or lifetime expiry, and the lifetime pauses while an attempt is in progress. The pairing window is now a static-code-only concept and its section moves under the Static Pairing Code Flow.

`client/pair-pending` now covers both cases, a gesture awaited or a dynamic attempt held back, and gains an optional `message`: a plain-text sentence for the operator of at most 200 characters, preferably in one of the server's `languages`, shown as device text with no markup interpreted. The server SHOULD surface the pending state and the message to the operator and apply its own timeout.

An abandoned attempt does not count against a window and counts as a failed dynamic attempt only if the code was already being emitted.


pairing.md
```diff
@@ -172,12 +172,12 @@ A failed key confirmation results in [`pair/abort`](#client--server-pairabort) w
...
+- **Counter.** The client maintains a single failure counter for the method. It is not partitioned by `server_id` or source IP, and does not persist across reboots.
 - **Increment.** The counter increments when the client starts emitting the pairing code, at most once per attempt. No other event increments it.
 - **Reset.** The counter resets to zero when the client's verification of `server_kc` succeeds, whether or not the attempt finalizes.
-- **Escalation.** When the counter reaches **5**, the method is **escalated**: every subsequent attempt is gesture-gated until a reset de-escalates it. Escalation is not an error state - the method stays offered.
+- **Backoff.** While the counter is **5** or higher, the method is in **backoff**: the client withholds [`client/pair-init`](#client--server-clientpair-init), signaling [`client/pair-pending`](#client--server-clientpair-pending), until a cooldown has elapsed since the last increment. Recommended cooldown: 1 minute when the counter reaches 5, doubling with each further failure, capped at 15 minutes. Backoff is not an error state - the method stays offered.
```
maximmaxim345: Was thinking about this not being required and essentially allowing a 1ms timeout client being Spec conformant, but that shouldn't be an issue from a security perspective at least right now: Implementations were already allowed to enable unpaired access and allowing the source role to be accessed without pairing.

This of course assumes we do not have a management surface (#183 is merged).

But this doesn't factor in future roles that might carry a higher security requirement:
At the very least for the implementation guide we need to strongly recommend to not do too lean cooldowns and list the scenario:
- Product shipped with insecure pairing timeouts, even though unpaired access is allowed this might not be a direct security problem right now.
- Sendspin adds a new Role with higher security requirements
- Products receives an update supporting this new role with the intention to gate it behind paired connections.
- While pairing timeouts can be updated, there is no way to detect attackers from legitimately paired servers, to keep security the device would essentially have to be factory reset

pairing.md
```diff
@@ -250,17 +250,14 @@ A decoder MUST reject malformed input, including a payload shorter than its vers
...
-
-An attempt is **gesture-gated** - the client withholds [`client/pair-init`](#client--server-clientpair-init) until a window is open - per the selected method's policy:
+The Static Pairing Code Flow gates every attempt on a **pairing window**: a state in which the client has decided to accept pairing attempts. The window admits up to **5** attempts and closes on a completed pairing, its fifth inner-authentication failure, drop of the connection carrying its attempts, operator cancellation, or window-lifetime expiry. An aborted or timed-out attempt only ends that attempt; the window stays open for another.
 
-- `static_pairing_code` - every attempt.
-- `dynamic_pairing_code` - only when the method is [escalated](#failure-counter).
+An attempt is **gesture-gated** - the client withholds [`client/pair-init`](#client--server-clientpair-init) until a window is open - for every `static_pairing_code` attempt. Dynamic Pairing Code attempts are never gesture-gated; their brute-force protection is the [failure counter's backoff](#failure-counter).
```
maximmaxim345: I already raised this concern once in: https://github.com/Sendspin/spec/pull/173#discussion_r3875298073.

The mitigation from that thread (incrementing on emission start instead of on a failed confirmation) only helped because escalation then capped it at 5 emissions. This PR removes the cap, so the rapid fire is back, just paced at 4 per hour instead of unlimited. Ten identical clips still give a guaranteed 12 seconds of attacker-chosen audio per attempt, so that is roughly 19 minutes a day, forever, on any idle speaker client.

This PR essentially adds a backdoor to bypass `unpaired_access`.

maximmaxim345 (commented):
Have two security concerns with this PR, left inline comments for both of them.

pairing.md
balloob: A client knows when they add a role and it requires them to adjust the cooldown.

pairing.md
balloob: Tracking this as https://github.com/orgs/Sendspin/projects/1?pane=issue&itemId=238792560. We should revert #173.

pairing.md
maximmaxim345: My argument is that adjusting the cooldown when adding a role is not enough (since it's already too late by then).
We need to note this down for the implementation guide, or write it out as a note or paragraph in this PR.

pairing.md
arturpragacz: That doesn't solve the problem. The problem is annoyance, not security. Annoyance is still a problem if it's saying digits on cool down constantly.

maximmaxim345: I disagree, CVEs in flac and opus decoders are unfortunately a thing. It's unfortunate but this considerably increases the attack surface. This is very likely to get noticed at a security audit if commercial devices want to implement Sendspin (just as leaving `unpaired_access` enabled by default, but they don't have to do that).

Also, 2s of audio are still enough for some script kiddy to force speakers to output NSFW words or sentences. This is IMO a lot more severe than just a couple dictated numbers.

pairing.md
maximmaxim345: Opened an issue to keep track of a future implementation guide #205

pairing.md
maximmaxim345: Annoyance is partly solved now with the hard limit of 20 (or it isn't an problem at all depending how the client handles pairing, they might require an action immediately).

pairing.md
```diff
@@ -175,14 +175,11 @@ SP:14DQ6FY7E4XTOP9HJ5LV6Z3PO57YPD4XT6T97N5Y
...
 
-#### Failure counter
+#### Failed attempts
 
-Brute-force protection for the Dynamic Pairing Code Flow is built around a failure counter that escalates the method to gesture-gating (see [Pairing Window](#pairing-window)). The following rules are mandatory for clients implementing `dynamic_pairing_code`:
+An attempt counts as failed once the client has started emitting the code and the attempt ends without a successful verification of `server_kc`. After 20 consecutive failed attempts the client MUST hold attempts back until a deliberate, manufacturer-defined operator action, such as a gesture as for a [pairing window](#pairing-window), an action in the manufacturer's own app, or a power cycle. The count is not partitioned by `server_id` or source address and resets on a successful verification or on that action. A client MAY hold attempts back earlier, by a cooldown or by an operator action, including from the first attempt.
```
arturpragacz: Power cycle is problematic as it's not obvious to users that it will do that, so it is not deliberate.

At the very least it should be limited to devices that are hardwired or otherwise designed to be stationary. I don't think it's a good idea to allow it for mobile speakers or some battery-operated devices, where turning the device on and off is common in its normal operation.

Also this is a repeated theme, but I'm not a fan of requiring manufacturers apps to do things. I don't think it should be explicitly mentioned in the standard.

pairing.md
```diff
@@ -255,18 +252,15 @@ A decoder MUST reject malformed input, including a payload shorter than its vers
 
 ### Pairing Window
```
arturpragacz: If it's only for static code flow now, then it should be a subsection there.

pairing.md
```diff
@@ -334,13 +328,15 @@ The pairing messages below are listed in the order they appear in the Dynamic Pa
...
+Reports that the client is holding back the selected attempt: no [pairing window](#pairing-window) is open, or its [attempt limit](#failed-attempts) has not admitted it yet. Sent immediately on receiving such a pairing [`server/activate`](messaging.md#server--client-serveractivate); [`client/pair-init`](#client--server-clientpair-init) follows once the client is ready. Does not start the [attempt](#entering-and-leaving-pairing) or its timeout. The server SHOULD surface the pending state to the operator and apply its own timeout (see [Entering and leaving pairing](#entering-and-leaving-pairing)); when `retry_after_ms` is present it SHOULD either outlast it or cancel the attempt and show the operator when to retry.
 
 - `pairing_index`: integer - see [Pairing index](#messages)
+- `retry_after_ms?`: integer - the client will be ready in this many milliseconds without operator action; absent when an operator action is needed
```
arturpragacz: With a hard stop at 20, it doesn't really meaningfully improve security, as it's the stop that is the limiting factor then.

I guess if implemented it does make the annoyance problem slightly more manageable.

On the other hand if it's too strict, it has the possibility of making the user experience more frustrating.

Also a nitpick, but do we really need ms precision here?

pairing.md
```diff
@@ -334,13 +328,15 @@ The pairing messages below are listed in the order they appear in the Dynamic Pa
...
+- `message?`: string - a short plain-text sentence for the operator, at most 200 characters, such as what to do to proceed, preferably in one of the server's [`languages`](messaging.md#server--client-serverhello). It comes from an unauthenticated peer: the server shows it as text attributed to the device, truncates it to that length, and MUST NOT interpret markup or links in it
```
arturpragacz, line 334: We could define this as an enum with all the ways that we envision this can be reasonably implemented.

That would both make translations easier and allow us to show some graphics, etc.

But it would be more work, and there's a risk that we miss some use cases.

arturpragacz (dismissed):

pairing.md
maximmaxim345: It is completely up to the manufacturer how the action will look. This is just a list of examples (it starts with "such as" after all). But I dropped it from the spec for now, I will re-add it with the PR for #205 later with more detail so the bad implementation example you listed shouldn't happen.

pairing.md
maximmaxim345: Done with 38c1401

pairing.md
maximmaxim345: Agreed, it was about security before my recent changes (I mean that we encouraged use of that with better UI for the timeout), but thats no longer required.

On the unit: we use ms in most places for consistency throughout the spec.

pairing.md
maximmaxim345: Yea, but I don't think we have enough information about this. We can think of ways that might happen (like popup on screen, press play button, hold play button, hold pairing button, hold power button, ...) but thats just too many possibilities to list out.
So we would need to at least group them to something simpler like "press button", but thats just not enough information to be useful.
We could add a `reason` and optional `message`, but I don't know how we might show that in a server UI... (since the message text is actually whats useful).

For translations, it's up to the client with the language sent through `server/hello`. (with highly specialized messages I don't think the server could even know about pre-translated versions for all cases)
Automatically translating it is an idea but I'm starting to overthink this problem.

maximmaxim345 (approved):

PR #195, opened 2026-09-01
Fold `stream/request-format` into `client/state`

### LINKED ISSUE #190: Move `stream/request-format` fields into `client/state` and drop the message
Follow-up to the alignment question in #177: `stream/request-format` carries dynamically changing values, which per the CONTRIBUTING rule belong in `client/state`. Looking closer, the whole message should go, because it is a request in name only.

A decline is invisible: when the server cannot honor a request, the stream continues and the request MUST NOT be treated as an error, so no message ever tells the client it was declined. The client learns the operative configuration exclusively from `stream/start`, which is exactly how state assertions would work. With no stream active the server SHOULD already remember the requested format for the next stream, so the request is de facto state, stored on the wrong side with a weaker guarantee. And no role lets a client request a format it cannot accept (player requests MUST come from `supported_formats`), so the fields always describe the client's current wishes.

Proposed shape: `player` gains an optional `format` preference in its state object (absent means the server picks per `supported_formats`); `artwork` moves the per-channel `source`/`format`/`width`/`height` into a state `channels` array, with `client/hello` shrinking to a channel count; `visualizer` moves `types`, `rate_max` and `spectrum` into state, with hello keeping only `buffer_capacity`. `stream/start` derives each role's configuration from the hello constants plus the role's current state, and answers a state change on an active stream with a new `stream/start` when the configuration changed.

This also collapses the last message that applied as a patch outside of management: after it, client self-description is always full state, and only management requests apply as patches with explicit outcomes.

After #182 even the management requests will be consistent (since no management messages will exist), making the protocol fully consistent in this area.

maximmaxim345:
`stream/request-format` was a request in name only: a decline is invisible (the stream continues and the request MUST NOT be treated as an error), so a client only ever learns the operative configuration from `stream/start`, and with no stream active the server already had to remember the requested format as de facto state. Per the CONTRIBUTING rule that fields expected to change during a connection belong in `client/state`, the message is dropped and its fields move there.

# Changes

`player` gains an optional `format` preference in its state object; absent means the server selects per the `supported_formats` priority order, a present preference must be one of the listed entries (which also removes the previously undefined partial-request case), and the server may still deviate, for example to match a track's native sample rate or to apply an operator-configured format. `artwork` moves the per-channel `source`/`format`/`width`/`height` into a state `channels` array and `artwork@v1_support` is removed entirely: the array is positional from channel 0, an uncovered channel is `source: 'none'` so clients may truncate after their last active channel, an array longer than 4 is a protocol error, and `format`/`width`/`height` are required only when `source` is not `'none'`. `visualizer` moves `types`, `rate_max` and `spectrum` into state; its hello object keeps only `buffer_capacity`. `stream/start` now derives each role's configuration from the hello constants plus the role's current state, answers a state change on an active stream with a new `stream/start` when the configuration changed, never starts a stream in response, and should wait for the state update a role (re-)activation requires before starting that role's stream.

# Breaking changes

The `stream/request-format` message is removed; clients that send it break, and servers drop the handler. Artwork clients no longer send `artwork@v1_support` and instead carry their channel configuration in `client/state`; visualizer clients keep only `buffer_capacity` in their support object and carry the rest in `client/state`. Players are affected only if they used `stream/request-format`; the new `format` state field is optional.

Conflicts with #177 in the player state object; whichever lands second needs a rebase.

Closes #190


messaging.md
```diff
@@ -349,6 +350,10 @@ Starts a stream for one or more roles. If sent for a role that already has an ac
 
 The server MUST NOT send `stream/start` to a client that is not [`available`](#client--server-clientstate) (e.g. a client whose output is taken by an [external source](#external-source-handling)).
 
+Each role's stream configuration is derived from what the client reports about itself: the role's support object in [`client/hello`](#client--server-clienthello) for constant capabilities, and the role's [`client/state`](#client--server-clientstate) object for the stream-configuration fields the client may change during the connection (each role may define both). After a role is added or re-added to `active_roles`, the server SHOULD wait for the [`client/state`](#client--server-clientstate) update the activation requires before starting that role's stream, so it does not start from stale state. When a `client/state` changes a role's stream-configuration fields while a stream is active for that role, the server re-derives the stream configuration and, if it changed, sends a `stream/start` with the new configuration. When no stream is active for the role, the server MUST NOT start one in response; the updated state applies to the next stream it starts for that role.
```
kahrendt: "After a role is added or re-added to `active_roles`, the server SHOULD wait for the `client/state`"
Artwork and visualizer streams now cannot start without a `client/state` message. Should this be a MUST instead?

roles/artwork/v1.md
```diff
@@ -3,39 +3,28 @@ This section describes messages specific to clients with the `artwork` role, whi
...
-
-The server MUST deliver each image at exactly the `width` by `height` declared for the channel. It MUST scale the source image to fit within those dimensions preserving its aspect ratio, MUST pad the remaining area with black, and MUST NOT crop the image.
+    - `format?`: 'jpeg' | 'png' - image format identifier. Required unless `source` is `'none'`
+    - `width?`: integer - width in pixels of the delivered image. Required unless `source` is `'none'`
+    - `height?`: integer - height in pixels of the delivered image. Required unless `source` is `'none'`
 
-**Note:** Clients can support 1-4 independent artwork channels depending on their display capabilities. The channel number is determined by array position: `channels[0]` is channel 0 (binary message type 8), `channels[1]` is channel 1 (binary message type 9), etc.
+The `channels` array is positional from channel 0 (length 1-4). A channel index the array does not cover is `source: 'none'`, so a client MAY truncate the array after its last active channel. An `artwork` object whose `channels` array is longer than 4 is a protocol error; the server should close the connection.
```
kahrendt: "the server SHOULD close the connection"

roles/visualizer/v1.md
```diff
@@ -10,21 +10,31 @@ Each visualizer binary message carries exactly one frame. The server emits messa
...
 
+When this configuration changes while a `visualizer` stream is active, the server re-derives the stream configuration and sends a [`stream/start`](../../messaging.md#server--client-streamstart) if it changed; with no active stream, the configuration applies to the next stream the server starts.
+
 ### Server → Client: `stream/start` visualizer object
 
 The `visualizer` object in [`stream/start`](../../messaging.md#server--client-streamstart) has this structure:
 
 - `visualizer`: object
-  - `types`: string[] - visualization data types the server will stream. MUST be a subset of the types the client requested (in [`client/hello`](#client--server-clienthello-visualizerv1-support-object) or the latest [`stream/request-format`](#client--server-streamrequest-format-visualizer-object))
+  - `types`: string[] - visualization data types the server will stream. MUST be a subset of the types the client requested in its current [`client/state`](#client--server-clientstate-visualizer-object)
```
kahrendt, line 39: In the artwork role, we specify that the server should close the connection on a malformed configuration. This has no parallel action to do if something is out of spec here.

roles/artwork/v1.md
```diff
@@ -3,39 +3,28 @@ This section describes messages specific to clients with the `artwork` role, whi
...
 
-### Client → Server: `stream/request-format` artwork object
-
-The `artwork` object in [`stream/request-format`](../../messaging.md#client--server-streamrequest-format) has this structure:
-
-Request the server to change the artwork format for a specific channel. The client can send multiple `stream/request-format` messages to change formats on different channels.
+The server MUST deliver each image at exactly the `width` by `height` declared for the channel. It MUST scale the source image to fit within those dimensions preserving its aspect ratio, MUST pad the remaining area with black, and MUST NOT crop the image.
 
-Response when an `artwork` stream is active: [`stream/start`](../../messaging.md#server--client-streamstart) with the new format, followed by immediate artwork updates through binary messages.
+When a channel's configuration changes while an `artwork` stream is active, the server responds with a [`stream/start`](../../messaging.md#server--client-streamstart) carrying the new configuration, followed by immediate artwork updates through binary messages.
```
kahrendt: Maybe be specific here that the artwork updates are sent for specific channels that had a configuration change (not all channels get re-sent).

messaging.md
```diff
@@ -190,7 +190,6 @@ Players that can output audio should have the role `player`.
   - `color@v1` - receives colors derived from the current audio
 - `player@v1_support?`: object - required if `player@v1` is listed, absent otherwise ([see player@v1 support object details](roles/player/v1.md#client--server-clienthello-playerv1-support-object))
 - `source@v1_support?`: object - required if `source@v1` is listed, absent otherwise ([see source@v1 support object details](roles/source/v1.md#client--server-clienthello-sourcev1-support-object))
-- `artwork@v1_support?`: object - required if `artwork@v1` is listed, absent otherwise ([see artwork@v1 support object details](roles/artwork/v1.md#client--server-clienthello-artworkv1-support-object))
```
kahrendt, line 193: How do we communicate the artwork role version to the server? (I may have missed it!)

maximmaxim345: Same as before: The client lists all their supported role versions in `client/hello.supported_roles`, after that the server acknowledges which of them is used in `server/activate.active_roles`.

roles/artwork/v1.md
maximmaxim345: Fixed in 9976569.

roles/artwork/v1.md
maximmaxim345: Fixed in 13b9077

messaging.md
maximmaxim345: Noticed that we never actually required clients to resend a role's state when the role becomes active, so a MUST wait could have deadlocked on re-activation. Pushed 21d9ac4 to fix that.

roles/visualizer/v1.md
maximmaxim345: Added in 912bc4a: 'spectrum' in types without a spectrum object is now a protocol error, matching the artwork rule.

We will later do a simplification round, we will then see if this is maybe not the final wording (but thats for later).

kahrendt (approved):

PR #204, opened 2026-09-02
Gate binary data on the role's `client/state` object

maximmaxim345:
Since role objects became optional per `client/state` message, a client can satisfy "has sent its initial `client/state`" with `available` alone and deliver the `player` object later on activation. The binary-data gate in `messaging.md` therefore opened before the server knew `output_delay_ms`, `required_lead_time_ms`, or `min_buffer_ms`.

The gate now requires the role's `client/state` object before that role's binary data. All built-in roles that receive binary data define one, so only application-specific roles without a state object fall back to the initial `client/state`. This matches the `stream/start` rule that waits for the activation update before starting a role's stream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,7 +543,7 @@ For synchronization, all timing is relative to the server's monotonic clock. The
 
 Client sends state updates to the server. Contains client-level state and role-specific state objects.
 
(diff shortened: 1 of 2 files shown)


kahrendt (approved):
Looks good!

PR #208, opened 2026-09-07
Let clients admit a pairing connection during playback

### LINKED ISSUE #197: Decide: let non-dynamic pairing proceed while another server is playing
Today a client admits one connection. `'playback'` outranks `'pairing'`, so an incoming pairing connection is rejected while another server plays, for every method.

Failure mode:

1. Music Assistant is playing on the speaker.
2. User adds Home Assistant. HA connects and sends `server/activate` with `activities: ['pairing']`.
3. Speaker replies `pair/abort` `concurrent_attempt` and closes.
4. HA shows "could not pair". Nothing says why or who holds the speaker.
5. User has to stop playback on every speaker, then retry.

Only pairing that emits the code through the playing output needs the speaker idle (#198). Pairing PSK, static code, and a dynamic code shown in an app or on a display do not.

Decide:

- Admit a pairing connection alongside the playback connection unless the client declares its emission takes the output (#198).
- Keep as is.

Cost of admitting: the client tracks one extra connection class, capped under the pairing-record limit.

maximmaxim345:
Today a client rejects a pairing connection while another server is playing, so the user must stop playback everywhere before adding a second server.

This PR makes it the client's choice. A client MAY hold one pairing connection next to the playback connection, or keep rejecting it with `pair/abort` `concurrent_attempt` as before.

Two rules keep the outcome defined: further incoming connections are arbitrated against the playback holder, and once the pairing connection drops `'pairing'` from its activities it is arbitrated as if newly incoming. The `concurrent_attempt` reason text now also covers a client that does not admit the second connection.

Closes #197


connection.md
```diff
@@ -31,12 +31,13 @@ A connection with empty `activities` ranks lowest.
...
 
 - A [pairing attempt](pairing.md#entering-and-leaving-pairing) is not displaced by an incoming `'playback'` or `'pairing'` connection.
 - When both the current holder and the incoming connection have empty `activities`, the incoming is admitted only if its `server_id` matches the last-playback server (and the existing one's does not); otherwise the existing is kept.
+- A client MAY admit one incoming `'pairing'` connection alongside the admitted `'playback'` connection, holding both. A client that does not rejects the incoming with [`pair/abort`](pairing.md#client--server-pairabort) reason `concurrent_attempt` and closes the connection. While both are held, further incoming connections are arbitrated against the `'playback'` holder. When a later `server/activate` drops `'pairing'` from that connection's `activities`, the client arbitrates it against the `'playback'` holder as if it were incoming.
```
kahrendt: The "A client that does not rejects" is a bit ambiguous, maybe specify if it not supported it should reject or something along those lines.

maximmaxim345: Changed it to "a client that does not admit both connections rejects the incoming ..." with 9d797bcf0a32.

kahrendt (approved):
Looks good, thanks!

PR #209, opened 2026-09-07
Add `client/leave` to replace the `available` flap

### LINKED ISSUE #202: Decide: replace the `available: false` / `available: true` flap with an explicit leave-group signal
Today a client leaves its group by sending `client/state` `available: false` immediately followed by `available: true`. The server's unavailable handling moves it to a solo group and does not rejoin it. The spec adds this is "only needed while the group's `playback_state` is `'playing'`".

A state field used as a verb. `available` already means "ready" and "output taken"; the flap adds "leave my group" as a transition.

Failure modes:

1. Server coalesces two `client/state` messages 5 ms apart (the player role allows coalescing timing updates). Flap lost, client stays in the group.
2. HDMI input toggles twice in a second. Client sends the flap for real reasons and is kicked to a solo group it did not ask for.
3. Client has to check `playback_state` before deciding whether to flap.

Decide:

- Add a `controller` command `leave` (solo group, stopped, `available` untouched). Ties it to the `controller` role.
- Add a role-independent client message for it so player-only clients can use it.
- Keep the flap and add: servers MUST NOT coalesce consecutive `client/state` messages that differ in `available`.

maximmaxim345:
A client that wanted to leave its group had to send `client/state` with `available: false` immediately followed by `available: true`.

This PR adds a role-independent `client/leave` message. The server handles it exactly as when the client becomes unavailable (solo group, playback stopped, previous group remembered for `switch`) while leaving `available` untouched. The flap text is replaced by a reference to the new message.

Closes #202

### DIFF
--- a/messaging.md
+++ b/messaging.md
@@ -282,7 +282,7 @@ A client's output can be taken over by a non-Sendspin activity (playing other me
 
 If the external activity can be interrupted by Sendspin playback at any time, the client SHOULD remain `available: true` so the server can take it over.
 
-To stop rendering its group's audio while performing non-Sendspin activity, a client MAY leave its current group by sending a `client/state` with `available: false` immediately followed by a `client/state` with `available: true` (the server behavior below then moves it to a solo group and does not rejoin it). This is only needed while the group's `playback_state` is `'playing'`.
+To stop taking part in its group's playback while performing non-Sendspin activity, a client MAY leave its current group with [`client/leave`](#client--server-clientleave). This is only needed while the group's `playback_state` is `'playing'`: a client in a stopped group keeps its grouping by staying, and later playback may still take it over.
 
 #### Non-interruptible activity (client becomes unavailable)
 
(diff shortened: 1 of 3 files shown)


messaging.md
```diff
@@ -312,6 +312,14 @@ Client sends commands to the server. Contains command objects based on the clien
 
 [Application-specific roles](README.md#application-specific-roles) may also include objects in this message (keys starting with `_`).
 
+### Client → Server: `client/leave`
+
+Leaves the client's current group. No payload fields.
+
+A client sends this when it no longer wants to take part in its group's playback, for example while performing an [interruptible non-Sendspin activity](#interruptible-activity-client-stays-available). `available` is unchanged.
```
kahrendt: "`available` is unchanged." Seems unnecessary to include here.

maximmaxim345: Fixed in eb7361f21b74

kahrendt (approved):
Thanks!

PR #212, opened 2026-09-07
Clarify player correction scope and minimum step

maximmaxim345:
The suggested sample-deletion/insertion formula can round to zero at low sample rates. State that the example is intended for 44.1 kHz and above, and clamp the calculated correction step to at least one frame. This does not restrict the sample rates players may advertise or change behavior in the intended range. The clamp improves behavior outside that range but does not provide complete low-rate support, since the existing per-chunk cap can still prevent correction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1280,7 +1280,7 @@ Each client is responsible for maintaining its own synchronization with the serv
 
 ### Suggested correction strategy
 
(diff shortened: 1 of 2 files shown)


kahrendt (approved):
Thanks!

PR #225, opened 2026-09-08
Define source readiness and restart authorization

maximmaxim345:
Source streaming is already gated by availability, but the spec does not require servers to wait for readiness before sending `start` or clearly reset start authorization when the role is removed. Require servers to wait for activation state and the latest received `available: true`, and require clients to ignore starts received while unavailable and wait for a fresh start after role reactivation. Servers tolerate chunks already in flight during removal. This uses existing messages and leaves the separate activation-state correlation gap unresolved.


kahrendt (approved):
Looks good!

Issue #226, opened 2026-09-08
Clarify the purpose of null role state


maximmaxim345:
`server/state` currently allows a role object to be `null`, which immediately clears its current state and any pending scheduled update. This is not restricted to role deactivation. The deactivation sequence also requires sending `null` before removing a state role from `active_roles`.

Is clearing state while keeping a role active useful in practice, or should `null` be reserved for the role-deactivation transition? For metadata and color, an otherwise empty object with a timestamp can already represent an empty snapshot. Controller state normally still exists even when playback is stopped.

We should decide whether `null` remains a general clear operation, whether its meaning should differ by role, and whether it is appropriate as the initial state after activation or reactivation. If we reserve it for deactivation, the rule must account for the current ordering: the clear arrives while the role is still active, before the activation update removes it.

This is a design question, not a proposal to change behavior yet.

PR #227, opened 2026-09-08
Send initial server state after role activation

maximmaxim345:
The spec constrains the first `server/state` message but does not require sending one after role activation. Require servers to promptly send current state or `null` when a server-state role is activated or reactivated, so controller, metadata, and color clients receive an initial state without waiting for a later change.

This uses the existing `null` clear-state semantics when no current state is available. Whether `null` should remain valid outside role deactivation is discussed separately in #226.


kahrendt (approved):
Looks good. We can always modify it further if necessary after making a decision about the null state.

PR #228, opened 2026-09-08
Clarify visualizer frame rates per periodic type

maximmaxim345:
The visualizer rate limit does not specify whether it applies per type or across all periodic messages. Define `rate_max` per periodic type in both client state and stream configuration, consistent with the existing aiosendspin v1 periodic scheduler. Clients should expect up to the declared rate for each selected periodic type, not that rate in total. Beat and peak remain exempt.


kahrendt (approved):
Thanks!

PR #229, opened 2026-09-08
Clarify visualizer timestamp ordering across stream updates

maximmaxim345:
The visualizer requires non-decreasing timestamps without specifying where ordering resets. Allow earlier timestamps after a visualizer clear or when starting a new stream, while preserving ordering across configuration updates to an active stream.

Note: Servers that reset timestamp ordering on configuration-only `stream/start` must preserve it or explicitly clear the visualizer stream before sending earlier timestamps.


kahrendt (approved):
Thanks!

PR #231, opened 2026-09-08
Clarify integer rounding of reported group volume

maximmaxim345:
The group volume field is an integer, but the average of eligible player volumes may be fractional. Clarify that the reported average is rounded to an integer.

Whether implementations round up or down is intentionally unspecified because the difference is too small to warrant standardizing. This does not change the per-player volume adjustment algorithm.


kahrendt (approved):
Good clarification, thanks!

Issue #230, opened 2026-09-08
Define player buffer-capacity accounting


maximmaxim345:
The player role describes `buffer_capacity` as a hard byte limit, but does not fully define how the server calculates occupancy. Its enforcement wording also says “should not” rather than `MUST NOT`.

Clarify which bytes count, including whether the audio message header is included and whether decoded or partially played chunks remain counted. Define when bytes become reusable, how `output_delay_ms` affects that timing, and how `stream/clear`, `stream/end`, and configuration updates affect accounting.

Unlike visualizer frames, an audio chunk’s timestamp identifies its first sample. Releasing the entire chunk at that timestamp could undercount audio still waiting to play. One option is to count each complete message until its last sample’s scheduled output, using a server-side estimate rather than client consumption acknowledgments.

Once the accounting model is defined, align the enforcement wording with the intended hard limit.

PR #232, opened 2026-09-08
Define and enforce visualizer buffer capacity

maximmaxim345:
Visualizer clients advertise `buffer_capacity`, but the spec does not require servers to enforce it or define when sent messages stop counting. Add a hard limit based on reassembled message bytes and future presentation timestamps on the server clock, with accounting reset by visualizer clear/end.

The player role already describes a hard limit, but its sending rule still uses lowercase "should not" rather than `MUST NOT`. Clarifying player accounting and aligning that wording are tracked separately in #230.


kahrendt (approved):
Thanks!

PR #233, opened 2026-09-08
Clarify output delay in the player sync target

maximmaxim345:
A player connected to an external amplifier with 100 ms of latency must output audio 100 ms before the intended audible timestamp. The existing sync-error definition compares that output with the unadjusted time-filter prediction, so a literal reading counts correct delay compensation as a 100 ms synchronization error. Clarify that the expected output time subtracts the configured `output_delay_ms`. Playback scheduling and the existing internal/external delay distinction are unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1249,7 +1249,7 @@ This section defines rules that require all implementations to provide a good ex
 
 ### Sync Accuracy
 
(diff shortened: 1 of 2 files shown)


kahrendt (approved):
LGTM

PR #235, opened 2026-09-08
Fix reactivation of roles without client state

maximmaxim345:
An application-specific streaming role with fixed capabilities and no `client/state` object can be disabled and re-enabled on the same connection. The client has already sent its initial `client/state`, but the stream-start rule asks for a fresh activation update even though there is no role object to send. Reuse the latest reported `available` for these roles and limit the fresh role-state gate to roles that define an object. Initial state reporting and the gates for stateful roles remain unchanged.


kahrendt (approved):
LGTM

PR #236, opened 2026-09-08
Clarify recovery after live buffer requirements increase

maximmaxim345:
A live group may be playing with 100 ms of lead when a timing update or joining player raises the requirement to 300 ms, but the additional live audio is not yet available. The spec requires recomputing send-ahead without defining the transition to a larger buffer. Allow players to temporarily fall below their requested `min_buffer_ms` and recommend that the server restore the requested minimum, subject to `buffer_capacity`, without prescribing a recovery mechanism or deadline.


kahrendt (approved):
Thanks

PR #237, opened 2026-09-09
Keep the dynamic pairing code stable across retries within an attempt

### LINKED ISSUE #201: Keep the dynamic pairing code stable across retries within an attempt
In the Dynamic Pairing Code Flow one attempt is one PAKE round. A wrong entry ends the attempt, and the retry draws new nonces, so the device shows a different code.

Failure mode:

1. Speaker says "four eight three, nine two one".
2. Operator types 483912.
3. `pair/abort` `pairing_code_mismatch`. Attempt over.
4. Server sends a new pairing `server/activate`. Speaker says "two zero five, six six eight".
5. Operator listens again and types again. Five of these and the method is in backoff.

The code should not change when the operator mistypes.

Proposed change: allow multiple PAKE rounds within one attempt, code fixed.

- After `pairing_code_mismatch` the attempt stays open. Server sends a fresh `server/pair-auth`, client answers `client/pair-auth`, round continues as before.
- `h`, `nonce_A`, `nonce_B`, `commit_B` unchanged, so the code is unchanged. Client keeps showing it or re-speaks it.
- `sid` gains a round counter: `"sendspin-pair-pake-v1" || h || counter || round` (uint32, from 0, +1 per `server/pair-auth`). `K_wrap` follows.
- Failure counter still increments per failed round. Brute-force cost unchanged.
- Attempt timeout still bounds the whole attempt.
- Either `pairing_code_mismatch` becomes "round failed, attempt continues", or add a `round_failed` reason. Open.

`nonce_B` is only revealed after `server_kc` verifies, so a failed round never exposes it. Static code already retries with a fixed code inside a window; this aligns the two flows.

arturpragacz:
Keep the dynamic pairing code stable across retries within an attempt.

Closes #201.

### DIFF
--- a/pairing.md
+++ b/pairing.md
@@ -10,7 +10,7 @@ This specification defines three pairing methods. Servers must implement all thr
 2. **Dynamic Pairing Code** - pairing with a per-session [Pairing Code](README.md#definitions) that the client derives from a commit-and-reveal binding to the Noise handshake and emits via an out-channel (display, speaker, etc.) for the operator to enter into the server. See [Dynamic Pairing Code Flow](#dynamic-pairing-code-flow).
 3. **Static Pairing Code** - pairing with a fixed [Pairing Code](README.md#definitions). Appropriate for devices with no out-channel; vulnerable to MITM if the pairing code is disclosed. See [Static Pairing Code Flow](#static-pairing-code-flow).
 
-A code-based pairing runs over a Sentinel-keyed connection: the channel is unauthenticated until the [PAKE](#pake) round completes. The round establishes trust from scratch and produces a new [long-term PSK](README.md#definitions).
+A code-based pairing runs over a Sentinel-keyed connection: the channel is unauthenticated until a [PAKE](#pake) round completes. The round establishes trust from scratch and produces a new [long-term PSK](README.md#definitions).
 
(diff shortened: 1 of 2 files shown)


pairing.md
```diff
@@ -370,6 +377,12 @@ Server's MCF tag, sent after the server has derived its CPace session key from `
 
 On receipt, the client verifies `server_kc` before sending [`client/pair-confirm`](#client--server-clientpair-confirm); see [Dynamic Pairing Code Flow](#dynamic-pairing-code-flow) / [Static Pairing Code Flow](#static-pairing-code-flow).
 
+#### Client → Server: `client/pair-retry`
+
+Asks for another [round](#rounds) of a [Dynamic Pairing Code Flow](#dynamic-pairing-code-flow) attempt. Sent in place of [`client/pair-confirm`](#client--server-clientpair-confirm) when `server_kc` fails to verify and the client admits another round.
```
```suggestion
Asks for another [round](#rounds) of a [Dynamic Pairing Code Flow](#dynamic-pairing-code-flow) attempt. Sent in place of [`client/pair-confirm`](#client--server-clientpair-confirm) when `server_kc` fails to verify and the client admits another round. The server SHOULD tell the operator that the pairing code did not match.
```
We also have UX guidance in `client/pair-pending`.

maximmaxim345: Dropped the "and prompt for the code again" from my suggestion since that will probably backfire with spoken codes depending on the interpretation...

maximmaxim345: Let's drop this suggestion. We didn't prescribe operator feedback for `pair/abort` either.

maximmaxim345 (approved):
Thanks @arturpragacz !

PR #241, opened 2026-09-09
Remove the server-supplied digit audio pack from dynamic pairing

### LINKED ISSUE #203: Revert https://github.com/Sendspin/spec/pull/173
We should not allow the server to submit audio clips (introduced in https://github.com/Sendspin/spec/pull/173)

It allows an attacker to play any kind of music on a speaker. Yes, it's only 2s, it might also be very uncomfortable. 

It also allows an attacker to attack the audio codecs used by the speaker, which becomes a huge attack surface.

chrisuthe:
Removes the digit audio pack that #173 introduced. During initial pairing the pack came from an unauthenticated peer, which thereby chose what the client's speaker played and drove the client's audio decoders before any trust was established. Speaking clients return to their own bundled voice, choosing the language against the server's `languages` hint in `server/hello` under RFC 4647 Lookup matching, as before #173.

Hand revert rather than `git revert`: #179, #194 and #237 have since rewritten the surrounding text, and the `languages` move from `server/activate` to `server/hello` in #173 stays, since `client/pair-pending`'s `message` field relies on it.

Changes, all in `messaging.md` and `pairing.md` with `README.md` regenerated:

- Binary message ID `2` returns to the reserved `2-3` range.
- `server/activate` `pairing.format?` no longer conditions `digits` on fitting a pack.
- Sequence diagram, Digits emission, Protocol Errors, the `dynamic_pairing_code` descriptor (`digit_audio` removed; `out_channels` unchanged), the Messages intro, `client/pair-init` and `server/pair-init` lose their pack references.
- The `Server → Client: Digit Audio Clip (Binary)` section is removed.

Closes #203


maximmaxim345 (approved):
Thanks @chrisuthe !

PR #242, opened 2026-09-09
Allow reconnecting to refresh unpaired access

maximmaxim345:
A client can enable unpaired access while an existing connection still carries the old setting from `client/hello`. Explicitly allow the client to send `client/goodbye` with reason `restart` and close that unpaired connection so the next hello advertises the updated setting. Refreshing remains optional and uses existing reconnect behavior without adding a message or requiring a device reboot.


kahrendt (approved):
LGTM

PR #247, opened 2026-09-10
Bind Pairing PSK finalize to its activation

maximmaxim345:
A delayed `client/pair-finalize` from a cancelled Pairing PSK attempt can be accepted as the response to a later activation, causing the peers to persist different keys. Require the existing indexed `client/pair-init` before Pairing PSK finalize and discard finalize messages while waiting for the matching init. The client sends both messages back-to-back, adding no round trip.

Finalize payloads and the empty server acknowledgment are unchanged. Ordered delivery keeps an old finalize before the next attempt's init, so it cannot be mistaken for that attempt's finalize.

## Breaking compatibility with older clients and servers

Updated servers require `client/pair-init` before Pairing PSK finalize, which older clients omit. Older servers treat the added init as a sequence violation in this flow. Both peers need the updated flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -759,15 +759,15 @@ Eviction needs no wire signal. An evicted server's next handshake references a `
 
 Pairing and playback are mutually exclusive on a connection. When a server moves an established connection into pairing it first quiesces the client's streams - sending [`stream/end`](#server--client-streamend) for active stream roles and a [`server/state`](#server--client-serverstate) with null role objects for state roles, as when a role is removed from `active_roles` - and then sends the pairing [`server/activate`](#server--client-serveractivate) with empty `active_roles`. The quiesce is stream-only: unlike an [`available: false`](#external-source-handling) transition, the client keeps its group membership and queued group state through the pairing activity - no move to a solo group, no previous-group memory, no bar on resuming in place.
 
-Each pairing `server/activate` admits one **pairing attempt**, in progress from its first pairing message - [`client/pair-init`](#client--server-clientpair-init) (pairing-code methods) or [`client/pair-finalize`](#client--server-clientpair-finalize) (Pairing PSK) - until success or [`pair/abort`](#client--server-pairabort). [`client/pair-pending`](#client--server-clientpair-pending) precedes an attempt and does not start it. The client bounds each attempt with an **attempt timeout** measured from its first message (recommended 2 minutes); on expiry it sends `pair/abort` with reason `attempt_timeout`.
+Each pairing `server/activate` admits one **pairing attempt**, in progress from [`client/pair-init`](#client--server-clientpair-init) until success or [`pair/abort`](#client--server-pairabort). [`client/pair-pending`](#client--server-clientpair-pending) precedes an attempt and does not start it. The client bounds each attempt with an **attempt timeout** measured from its first message (recommended 2 minutes); on expiry it sends `pair/abort` with reason `attempt_timeout`.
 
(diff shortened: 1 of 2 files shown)


arturpragacz (changes requested):
I'm not a fan of extending `pairing_index` to yet another message.

Could we instead just add an extra `client/pair-init` to the Pairing PSK Flow?

pairing.md
```diff
@@ -32,15 +32,15 @@ Eviction needs no wire signal. An evicted server's next handshake references a `
...
 The same `server/activate` can also end a pairing attempt without finalizing: sent in place of [`server/pair-finalize`](#server--client-serverpair-finalize), it persists nothing and discards any received PSK. A client that, after sending [`client/pair-finalize`](#client--server-clientpair-finalize), receives `server/activate` likewise persists nothing.
 
-After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`.
+After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`. In the [Pairing PSK Flow](#pairing-psk-flow), the server MUST silently discard `client/pair-finalize` messages received before a `client/pair-init` with a matching `pairing_index`.
```
arturpragacz, line 41:
```suggestion
After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`.
```
I think that already follows from the general language we have.

README.md
```diff
@@ -759,15 +759,15 @@ Eviction needs no wire signal. An evicted server's next handshake references a `
...
 The same `server/activate` can also end a pairing attempt without finalizing: sent in place of [`server/pair-finalize`](#server--client-serverpair-finalize), it persists nothing and discards any received PSK. A client that, after sending [`client/pair-finalize`](#client--server-clientpair-finalize), receives `server/activate` likewise persists nothing.
 
-After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`.
+After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`. In the [Pairing PSK Flow](#pairing-psk-flow), the server MUST silently discard `client/pair-finalize` messages received before a `client/pair-init` with a matching `pairing_index`.
```
arturpragacz, line 768:
```suggestion
After sending `pair/abort` or leaving pairing, a server silently discards pairing messages still in flight from the client - messages sent before the client observed the abort or leave `server/activate`. A client that has aborted an attempt likewise silently discards pairing messages received before the next `server/activate`.
```

arturpragacz (changes requested):

README.md
chrisuthe: also already understood from the language included.

arturpragacz (approved):

PR #257, opened 2026-09-10
Clarify player and visualizer capability values

maximmaxim345:
Require at least one supported format for an advertised player and explicitly allow zero startup-lead and ongoing-buffer requests through non-negative timing fields. Allow an empty visualizer type list, while requiring a positive rate cap. Existing synchronization requirements remain unchanged.

This tightens the contract for clients that advertise no player formats or report a zero visualizer rate cap, including with an empty or event-only type list. Zero buffer capacity, spectrum bounds, empty artwork channels, and resource maxima are outside this change.


kahrendt (approved):
LGTM

PR #259, opened 2026-09-10
Distinguish WebSocket messages from frames

maximmaxim345:
The spec calls Noise ciphertexts and handshake JSON WebSocket frames in some places, which can be read as requiring one RFC 6455 frame per message. Specify complete WebSocket messages as the transport boundary, with WebSocket reassembly before protocol processing and separate Sendspin fragmentation. Align the re-handshake and fragmentation wording while retaining control, audio, and visualization frames and leaving wire formats and limits unchanged. Keep qualifiers only where they distinguish layers.


chrisuthe:
Heads-up: I pushed two small commits onto this branch, so please pull before pushing again rather than force-pushing over them.

- 560c18c drops "reassembled by the WebSocket layer" and says Sendspin operates only on complete WebSocket messages. On ESP-IDF the client library hands over individual frames, and the C++ client does the reassembly itself, so the spec should state the boundary as a requirement on the implementation rather than a property of the transport.
- 9344851 uses "Noise transport message" for the encrypted unit everywhere in messaging.md. The fragmentation section already used that term; the three other sentences said "Noise transport ciphertext" for the same object.

The rest of the change looks right to me, including keeping "frame" for RFC 6455 control frames and PCM/visualizer frames.

PR #262, opened 2026-09-10
Clarify that `available` covers the whole client

maximmaxim345:
Availability was described in terms of the client's output, which left a source-only client without a way to report that its input was taken over. Reword `client/state` and External Source Handling so the client, not its output, is the subject, and state in the `false` bullet that the client will not yield to Sendspin on request, since an interruptible activity also uses the client while it stays available. Align the source-role note to match. No change to whole-client gating, group effects, or source stop and restart rules, and no per-role availability.


kahrendt (approved):
Good distinction, thanks!

PR #265, opened 2026-09-11
Report init failures with a cleartext `server/error`

### LINKED ISSUE #200: Decide: give non-cryptographic failures a visible reason before closing
Today every handshake-phase failure and pairing protocol error closes the WebSocket with no application-level message. There is no diagnostic channel anywhere.

Failure mode:

1. New client build sends `client/init` with `version: 2`.
2. Server closes the socket.
3. Client log: "disconnected". Server log: "disconnected".
4. Same symptom as a bad clock, a corrupt record, or an attacker probing the port.

Silent close is right for AEAD and PAKE failures. Version, suite, and malformed cleartext are not secrets.

Decide:

- Send a WebSocket close code and reason (RFC 6455 close frame) for the non-crypto cases: `unsupported_version`, `unsupported_suite`, `malformed`, `protocol_error`. No new message type; works before and after the handshake.
- A cleartext `server/error` during the `client/init` / `server/init` phase only.
- Keep silent close everywhere.

Caveat for either reason option: pre-handshake messages are unauthenticated, so a reason is a hint, and the list must never grow to include crypto outcomes.

kahrendt:
Resolves #200.

Every handshake-phase failure currently closes the connection with no application-level message. A client sending `version: 2` sees the same "disconnected" as a bad clock, a corrupt record, or a probe.

This adds a cleartext `server/error` message, sent by the server in place of `server/init`, with reasons `unsupported_version`, `unsupported_suite`, and `malformed`. All three describe `client/init` alone. Every other handshake-phase failure stays silent: a client-side rejection of `server/init`, handshake timeout, malformed inner Noise payload, `psk_id` miss, AEAD failure, and a cleartext frame received after switching to transport mode. The last one keeps a forged `server/error` from being honored once the channel is encrypted. The reason is unauthenticated and is a hint for logging and operator display.

**Why `protocol_error` is not a reason.** Pairing protocol errors mix benign cases (a missing field, a wrong-length CPace share) with PAKE outcomes (a `wrapped_nonce_B` that fails to decrypt), which the spec closes silently rather than reporting as `pairing_code_mismatch`. A reason on the benign cases would let a code-guessing attacker distinguish them by whether a reason arrives. A single reason for all of them adds nothing over the close. Pairing errors also occur inside the encrypted channel, after a peer has already completed the handshake, so the diagnostic value is low.

### DIFF
--- a/messaging.md
+++ b/messaging.md
@@ -13,7 +13,7 @@ Once the WebSocket connection is established, Client and Server perform an initi
 
 No other messages should be sent before the initial [`server/activate`](#server--client-serveractivate) arrives, except possibly [`client/goodbye`](#client--server-clientgoodbye). See [Encryption](connection.md#encryption) for cryptographic details.
 
-Cleartext handshake messages (`client/init`, `server/init`, `noise/handshake`) are sent as WebSocket **text** frames containing JSON. After the encrypted channel is established, all messages are sent as WebSocket **binary** frames carrying Noise transport ciphertexts.
+Cleartext handshake messages (`client/init`, `server/init`, `noise/handshake`, `server/error`) are sent as WebSocket **text** frames containing JSON. After the encrypted channel is established, all messages are sent as WebSocket **binary** frames carrying Noise transport ciphertexts.
 
 WebSocket control frames (Ping, Pong, Close; RFC 6455) are not Sendspin messages: they remain valid at any time, are not encrypted at the Noise layer, and Ping/Pong is the expected connection-liveness mechanism.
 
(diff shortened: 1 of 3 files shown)


messaging.md
```diff
@@ -154,12 +154,21 @@ The encrypted payload carried inside each Noise handshake message is a UTF-8 JSO
...
 The same `noise/handshake` message is used for the in-band [re-handshake](connection.md#re-handshake): the two messages then travel as ordinary encrypted JSON messages (binary frames, message type `0`), not bare Noise bytes. Noise message 2 is still encrypted under the pre-re-handshake transport keys; the first frame each side sends after the handshake completes uses the new keys.
 
+### Server → Client: `server/error`
+
+Sent by the server in place of [`server/init`](#server--client-serverinit) when it cannot accept the client's [`client/init`](#client--server-clientinit). The server closes the connection after sending. See [Failure Handling](connection.md#failure-handling).
+
+- `reason`: string - one of:
+  - `unsupported_version` - the client's `version` is not one the server implements
+  - `unsupported_suite` - the client's `suite` is not one the server implements
+  - `malformed` - `client/init` is not valid JSON of the defined shape
```
maximmaxim345, line 170: If we view it very strictly, a higher version `version` might make the JSON invalid. Some implementations will probably parse the message as a whole, so a regular `malformed` error might come up before the `version` field is ever parsed.

messaging.md
```diff
@@ -154,12 +154,21 @@ The encrypted payload carried inside each Noise handshake message is a UTF-8 JSO
...
+A malformed inner handshake payload (not valid UTF-8 JSON of the shape above) is a [silent failure](connection.md#failure-handling) and closes the WebSocket.
 
 After both handshake messages have been exchanged, both sides switch to Noise transport mode (all subsequent messages travel as the binary Noise-ciphertext frames described above).
 
 The same `noise/handshake` message is used for the in-band [re-handshake](connection.md#re-handshake): the two messages then travel as ordinary encrypted JSON messages (binary frames, message type `0`), not bare Noise bytes. Noise message 2 is still encrypted under the pre-re-handshake transport keys; the first frame each side sends after the handshake completes uses the new keys.
 
+### Server → Client: `server/error`
+
+Sent by the server in place of [`server/init`](#server--client-serverinit) when it cannot accept the client's [`client/init`](#client--server-clientinit). The server closes the connection after sending. See [Failure Handling](connection.md#failure-handling).
```
maximmaxim345, line 165: The handshake sequence at the beginning of `messaging.md` only lists `client/goodbye` as an exception, can you add `server/error` there too.

messaging.md
kahrendt: I gave it an order: validate `version` first, then the shape for `malformed`, and finally evaluate `suite`.

maximmaxim345 (approved):
Thanks!

PR #266, opened 2026-09-11
Clarify artwork stream lifecycle and data handling

maximmaxim345:
Clarify artwork receiver behavior and stream-end effects. Changes beyond keyword normalization:

- Require artwork `stream/end` to clear the current image as well as discard any pending image, so the channel stops displaying artwork.
- Replace ambiguous binary-message rejection guidance with a sender prohibition outside an active artwork stream. During an active stream, unavailable clients SHOULD discard otherwise valid image data and MUST NOT close solely because it arrived - availability updates and image data can cross in flight.
- Require clients discarding artwork during an active stream to process announces and cancels and count each part's `data` bytes toward `total_size` - discarding data must not lose track of transfer boundaries.
- Separate malformed artwork messages from malformed transfer sequences. Message-format errors still require closure, while the transfer-sequence closure rules apply within an active artwork stream - this avoids mandating closure solely for valid data arriving outside a stream.

Addresses part of #100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1591,7 +1591,11 @@ Each channel's configuration MUST match the channel's current [`client/state`](#
 
 ### Server → Client: Artwork (Binary)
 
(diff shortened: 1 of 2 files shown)


kahrendt (approved):
Thanks!

PR #269, opened 2026-09-11
Clarify state publication for scheduled metadata updates

maximmaxim345:
Align metadata state publication with scheduled updates and progress extrapolation. Changes beyond keyword normalization:

- Exempt playback-state changes already conveyed by scheduled metadata updates from requiring another metadata state message - the client already has the change and its effective time.
- Apply the shared reporting requirement to active roles' `server/state` objects. Scheduled updates taking effect and playback progress advancing as reported require no new message - clients can advance progress locally without repeated server updates.

Addresses part of #100.


kahrendt (approved):
Good clarification here!

PR #270, opened 2026-09-11
Clarify player buffering and synchronization requirements

maximmaxim345:
Normalize player requirement wording and clarify which rules are mandatory. Changes beyond keyword normalization:

- Change `stream/clear` handling from a recommendation to a MUST: clients clear all buffered audio chunks and continue with chunks received after the message - seeks must not replay buffered audio from the old position.
- Change byte-cap enforcement from a recommendation to a MUST NOT: servers cannot send data that would exceed the player's `buffer_capacity` - the advertised capacity is a hard limit, not a buffering target.
- Tie startup timestamps to the specified minimum send-ahead rather than ambiguously requiring the player's lead-time hint - the spec explicitly allows servers to provide less than `required_lead_time_ms`.
- Require state updates when a player state field changes, and apply timing debounce to measurements before changing the reported fields - debouncing must not delay reporting a field that has already changed.
- Clarify that player state is sent only while the role is active. Check outgoing commands against the latest player state received by the server, while clients ignore commands absent from their current supported-command list - support changes and commands can cross in flight.
- Replace ambiguous binary-message rejection guidance with a sender prohibition outside an active stream. During an active stream, unavailable clients SHOULD discard otherwise valid audio and MUST NOT close solely because it arrived - availability updates and audio can cross in flight.
- Remove the example's separate mandatory per-chunk correction bound and refer its one-shot corrections to the existing rarity requirement. The protocol-wide correction-quality requirements remain unchanged - the speed cap is measured over 150 ms, not per chunk.

Addresses part of #100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1172,9 +1172,9 @@ For each [`stream/start`](#server--client-streamstart) the server SHOULD select 
 
 The `player` object in [`client/state`](#client--server-clientstate) has this structure:
 
-Informs the server of player-specific state changes. Only for clients with the `player` role.
+Informs the server of player-specific state changes. Only for clients whose `player` role is active.
 
-State updates must be sent whenever any state changes, including when the volume was changed through a `server/command` or via device controls.
+State updates MUST be sent whenever a field in the `player` state object changes, including when the volume was changed through a `server/command` or via device controls.
 
 - `player`: object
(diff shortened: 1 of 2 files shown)


kahrendt (approved):
Thanks!

PR #271, opened 2026-09-11
Clarify visualizer buffering and data handling requirements

maximmaxim345:
Normalize visualizer requirement wording and clarify stream handling. Changes beyond keyword normalization:

- Change `stream/clear` handling from a recommendation to a MUST: clients clear all buffered visualization data and continue with data received after the message - seeks must not display buffered frames from the old position.
- Replace ambiguous binary-message rejection guidance with a sender prohibition outside an active visualizer stream - "reject" did not specify whether to discard data or close the connection.
- During an active stream, unavailable clients SHOULD discard otherwise valid visualization data and MUST NOT close solely because it arrived - availability updates and visualization data can cross in flight.
- Remove the recommendation to set `rate_max` to the display refresh rate - the desired visualization update rate need not match the display refresh rate.

Addresses part of #100.


kahrendt (approved):
LGTM, nice refinements

PR #272, opened 2026-09-11
Do not quiesce playback for pairing

### LINKED ISSUE #198: Decide: quiesce playback on entering pairing only for dynamic_pairing_code
Today entering pairing on an established connection quiesces every stream (`stream/end`, null state objects) before the pairing `server/activate`, whatever the method.

Failure mode:

1. Speaker is playing on an unpaired-access session.
2. Operator upgrades it to paired with `pairing_psk`.
3. Music stops for a re-handshake and two JSON messages that never touch the speaker.

Whether pairing needs the output is a client property, not a method property. A dynamic code shown in the vendor app or on a display does not touch audio; only emission through the playing speaker does. The server cannot know; the client can.

Decide:

- Client declares it: a boolean in the `dynamic_pairing_code` descriptor (or `out_channels` containing `'speaker'` means it). Server quiesces only then. Every other pairing keeps `'playback'` and `active_roles` live.
- Client handles it locally: no quiesce on the wire; a client that needs its output pauses its own rendering during the attempt.
- Keep as is.

Blockers for the first option: the allowed activity sets are single-purpose today, and the pairing PSK flow re-handshakes to a PSK on which `'playback'` is not allowed. Interacts with #185.

arturpragacz:
Do not quiesce playback for pairing.

Keep the connection state during pairing.

Closes #198.


kahrendt (approved):
Thanks!

PR #273, opened 2026-09-16
Require fresh source starts after input streams end

maximmaxim345:
Clarify source start authorization, stopping, and receiver behavior. Changes beyond keyword normalization:

- Change `stop` handling from a recommendation to a MUST: cancel pending startup, stop sending audio, and send `client-stream/end` if an input stream is open - stopping before startup completes must prevent the pending opening too.
- Limit each `start` to one opening and require a newly received server `start` before reopening after `client-stream/end`. Repeated starts while starting or streaming cannot accumulate permission for later openings.
- Explicitly require cancellation of pending startup and stopping on role removal or unavailability - returning to an active, available state must not revive an old request.
- Express initial and reactivation authorization in terms of receiving a server `start`, not merely the server sending it. Distinguish opening an input stream from an in-place format update.
- Prohibit treating an otherwise valid opening as unsolicited solely because the server sent its `start` before receiving the preceding end or availability update - messages in opposite directions can cross, so that command may still be fresh when the client receives it.
- Replace ambiguous binary-message rejection guidance with mandatory connection closure for source audio without an open input stream or after the latest received client state reports unavailability.
- Clarify that sending `stop` or removing the source role does not itself close the input stream at the server. Tolerate valid in-flight responses, including a delayed `client-stream/start`, until `client-stream/end` arrives, and process start/end messages in received order even when discarding audio.

Addresses part of #100.


kahrendt (approved):

kahrendt:
Thanks!

PR #274, opened 2026-09-16
Clarify activation and shared stream lifecycle requirements

### LINKED ISSUE #100: Tighter Key Word usage

maximmaxim345:
Complete the BCP 14 boilerplate, normalize requirement wording, and move operative text out of non-normative notes. Clarify in CONTRIBUTING that requirements belong outside notes regardless of capitalization. Changes beyond keyword normalization:

- Make the initial activation message gate mandatory. The server's gate ends when it sends activation, and the client's ends when it receives activation. Preserve `server/error` for init failures and permit encrypted `client/goodbye` after the initial Noise handshake without changing silent-failure handling.
- Base client role eligibility on received activation state. Require servers to ignore inactive-role objects in `client/state` and `client/command` without closing solely for their presence, while still processing client-level fields and active-role objects - the client may not yet have received the role removal.
- Require lifecycle messages to be processed in delivery order even when role data is discarded. Discarding payloads cannot bypass Noise processing, fragment reassembly, or required message checks.
- Clarify that local unavailability does not end a downstream stream and that clients must continue processing its lifecycle messages. Base the server's `stream/start` availability check on the latest client state it received.
- Make buffer clearing on `stream/clear` explicit and change `stream/end` output stopping and buffer clearing from a recommendation to a MUST.
- Restrict `stream/clear` to active streams: an explicit role list must be non-empty and contain only roles with active streams. Omitting `roles` clears all active player and visualizer streams, and the message cannot be sent when no targeted streams are active - this keeps clear and end targeting consistent without requiring both roles to be streaming.
- Restrict `stream/end` to active streams: an explicit role list must be non-empty and contain only roles with active streams, and the message cannot be sent when none are active. Align role-removal and unavailability teardown with that restriction.
- Include sources as well as players in the clock-synchronization prerequisite for reporting availability.
- Define support-object keys in both hello messages as the role-version identifier followed by `_support`, when that role version defines a support object.
- Make `server/unpair` subject to initial and re-handshake sequencing, and prevent the generic permission to omit `client/goodbye` from overriding cases where it is required.

This is the final PR for addressing the issues I found with #100.

Closes #100.


kahrendt (approved):
Very thorough, thank you!

PR #275, opened 2026-09-16
Clear role state on deactivation

### LINKED ISSUE #226: Clarify the purpose of null role state

maximmaxim345:
Role deactivation currently requires a separate `server/state` message with null role objects. Make clients clear current and pending state when a role is removed from `active_roles` instead, and remove nullable role objects from `server/state`. Servers must not activate a state role until they can provide its complete state, and must deactivate it if that becomes impossible. Metadata and color use timestamp-only empty snapshots when there is no content, while active controllers must provide complete state.

Closes #226


kahrendt:
@chrisuthe Could you double check my merge commit and fix are logical/correct?

kahrendt (approved):

maximmaxim345 (commented):
Merge looks good, thanks @kahrendt !

PR #282, opened 2026-09-16
Clarify visualizer's rate_max applies only to periodic events

### LINKED ISSUE #278: Make the `peak` throttling statement normative in the visualizer role
`roles/visualizer/v1.md` opens with "`beat` and `peak` are event-driven and not throttled by `rate_max`; all other types are periodic", and the `client/state` `rate_max` field says it "applies independently to `loudness`, `f_peak`, `spectrum`".

Neither sentence uses an RFC 2119 keyword, so it is unclear whether a server that detects onsets only at its periodic analysis hop (so `peak` events are effectively rate-limited to `rate_max`) is conformant. The reference implementation currently does exactly that. Suggest either "a server MUST NOT throttle `beat` or `peak` by `rate_max`" or an explicit statement that onset detection resolution is implementation-defined.

kahrendt:
- Clarifies that `rate_max` applies only to periodic types, so how often a server detects `beat` or `peak` events is outside its scope
- Updates the `rate_max` field description to name both `beat` and `peak` as unaffected; it previously mentioned only `beat`

Closes #278


chrisuthe:
Nice

PR #283, opened 2026-09-16
Clarify buffer handling on `stream/start` messages during an active stream

### LINKED ISSUE #276: Clarify buffer handling when an in-place `stream/start` changes the player codec or sample rate
`messaging.md` ("Server → Client: `stream/start`") says a `stream/start` for a role with an active stream "updates the stream configuration without clearing buffers", and that if a parameter change requires rebuffering the receiver "handles this internally: it does not clear buffers unless its implementation requires it".

For a change of `codec`, `sample_rate`, `channels` or `bit_depth` on the player role this leaves the timeline ambiguous. The already-buffered chunks are in the old format and run to their own tail; chunks after the new `stream/start` are in the new format. A client that follows the text literally holds both and must decode two formats against one timeline. A server that assumes the client flushes on the new `stream/start` (aiosendspin does, and re-sends from `now + lead` in the new format) produces a gap for a client that does not flush, and a client that flushes produces a gap for a server that continues the old timeline.

Suggested clarification: state which side owns the transition. Either (a) the server MUST continue the existing timeline and MAY only change format at a chunk boundary such that the first new-format chunk's timestamp follows the last old-format chunk, and the client keeps both queued; or (b) a format change on an active stream is treated like `stream/clear` for the player role: the client drops buffered audio and the server re-applies the startup lead. Option (b) matches "Track transitions ... `stream/start` to update the existing stream configuration" only if gapless is not expected across a format change.

Reference implementation behaviour today: aiosendspin `server/push_stream.py` `on_role_format_changed` and `server/roles/player/v1.py` `_begin_format_transition` assume (b).

kahrendt:
`stream/start` on an active stream let the client clear buffers "unless its
implementation requires it", leaving the transition timeline ambiguous when
the player format changes. The player role already assumes the timeline
continues, so state that explicitly.

- messaging.md: an in-place `stream/start` updates the configuration without
  ending the stream. Each role defines how previously received data is
  handled, since artwork discards its pending image while the player keeps
  its chunks.
- roles/player/v1.md: the format changes at a chunk boundary. Servers MUST
  continue the timeline without resending audio. Clients MUST decode each
  buffered chunk in the format in effect when it arrived. A player that
  cannot switch formats gaplessly SHOULD list a single `sample_rate` and
  `channels` and let the server resample.

Fixes #276

### Breaking changes

- **aiosendspin server** (since #336), **sendspin-python-cli**, **SendspinDroid**: flush on an in-place `stream/start`. Must continue the timeline instead.
- **sendspin-go**, **sendspin-rs** example, **aiosendspin client**: swap the decoder immediately, misdecoding queued old-format chunks. Must apply the new format per chunk.

sendspin-js, SendspinKit, and sendspin-cpp already conform. The conformance suite needs a scenario for a format change with chunks in flight.


chrisuthe:
Perfect, this is a good clarification.

Issue #205, opened 2026-09-03
Add implementation guide for (sdk) developers on the spec repository


maximmaxim345:
The spec should be joined by a non-normative implementation guide, and this issue proposes creating one and tracks what goes in it. Reviewers have deferred guidance to such a guide nine times so far, and because the guide does not exist, eight of those nine items survive only in closed review threads.

The split is clear in review: the spec says what goes on the wire, and the deferred items say how to build a client or server well. Each item below was kept out of the spec for that reason, not because anyone disagreed with it.

Add new items to this issue rather than opening separate ones, and link here from any thread that defers something.

### Pairing method choice

A client with a usable out-channel should offer `dynamic_pairing_code` rather than `static_pairing_code`, and should offer `qr_code` when its display can render one. This is the one item whose text is still in the spec: `pairing.md` holds it with a lowercase `should` until the guide exists (#189, #184). Move it here at that point.

A client offers at most one pairing-code method because offering the user both is too complicated to explain and the difference between them is not something a user should have to understand (#186).

### Timing and security

Pairing cooldowns must not be set too low, and the guide needs to explain what a low cooldown costs later. #194 requires a client to hold attempts back after 20 consecutive failures until an operator acts, and leaves any earlier cooldown to the client as a `MAY`, so the guide is where a client learns what to choose under that `MAY`. A product ships with a permissive cooldown, Sendspin adds a role with higher security requirements, the product gates that role behind pairing, and the operator then cannot tell an attacker's pairing record from a legitimate one, so only a factory reset restores security. Raising the cooldown when the role is added is already too late (#194).

The operator action that ends the 20-failure hold must be deliberate for that product, and the spec names no examples on purpose. Artur's test: if a reboot happens naturally during normal operation, it ceases to be a meaningful signal. So a plain power cycle only fits hardwired or otherwise stationary devices that nobody unplugs by accident, not devices on switched outlets or smart plug schedules, and not portable or battery devices. Those use a power-cycle pattern, a long press on a control, or the device's own UI. Whatever is chosen is reported in `reason` and described in `message` on `client/pair-pending` (#194).

A client needs a method for computing `min_buffer_ms`. The spec requires the field and gives no way to derive a value, unlike `required_lead_time_ms`. One suggested starting point is the distribution of time-message latencies at roughly the 95th percentile, pending experiments (#69).

[...]

PR #285, opened 2026-09-16
Move one pairing code method cap

arturpragacz:
Move the one-pairing-code-method cap to implementation guide.

#189 capped clients at one pairing-code method and added a server rule for advertisements that list both. #186 recorded that the cap is not a wire change: `supported_pair_methods` carries whatever the client offers, and `server/activate` names one method. The reason for the cap was UX, offering users both options is too confusing, which is a product decision of the same kind #189 moved out of the spec for method choice. #205 already lists it under the guide's "Pairing method choice" entry.

This relocates the rule rather than reversing it. Clients should still offer one pairing-code method; that guidance, plus "a server offered both prefers `dynamic_pairing_code`", goes to the implementation guide.


pairing.md
```diff
@@ -2,7 +2,7 @@
 
 Pairing is the one-time setup that mutually authenticates a client and a server. The pairing flow uses the same WebSocket endpoint and [`KKpsk2`](connection.md#encryption) Noise pattern as every other connection; only the PSK fed into the handshake and the client's post-handshake routing differ (see [Pre-Shared Key](connection.md#pre-shared-key)). After any successful pairing both sides persist the new pairing record, then the server initiates an in-band [re-handshake](connection.md#re-handshake) to the newly delivered `long_term_psk`, promoting the channel to a paired session without closing the WebSocket.
 
-This specification defines three pairing methods. Servers MUST implement all three; clients MUST implement Pairing PSK and MAY additionally offer at most one pairing-code method: Static Pairing Code or Dynamic Pairing Code.
+This specification defines three pairing methods. Servers MUST implement all three; clients MUST implement Pairing PSK.
```
marcelveldt:
```suggestion
This specification defines three pairing methods. Servers MUST implement all three; clients MUST implement Pairing PSK and MAY implement additionally other methods.
```

marcelveldt (approved):

PR #238, opened 2026-09-09
Make Opus optional for servers and warn about its license fees

marcelveldt:
Patent claims asserted as essential to Opus are licensed through a pool that charges per device, and the pool has started litigating. Sendspin does not need Opus on a LAN: FLAC and PCM are unencumbered and the bandwidth saving rarely matters there.

- servers MUST support `flac` and `pcm` and MAY support `opus`; clients MUST list at least one `flac` or `pcm` format and sources MUST be able to produce one, so any client works with any server
- servers that support the source role list the codecs they accept from sources in a new `source@v1_support` object in `server/hello`
- the player and source roles note that Opus is patent-encumbered and that commercial implementers are responsible for any license fees

PR #286, opened 2026-09-16
Clarify player format selection

maximmaxim345:
Follow up fix for a tiny inconsistency introduced with #238.

The `format?` description summarizes priority ordering without mentioning server capabilities or permitted overrides. Replace that summary with a link to the existing format-selection rules, which cover unsupported preferences such as Opus on a server that cannot produce it. This is a wording clarification with no protocol behavior change.


kahrendt (approved):
Thanks!

PR #287, opened 2026-09-16
Stop repeating hellos during re-handshake

### LINKED ISSUE #47: Decide what state survives re-handshake
Re-handshake keeps the WebSocket open but repeats `server/hello`, `client/hello`, and `server/activate`, without defining which state survives, including time synchronization, active streams, buffered data, and role state. Decide whether the repeated hello/activation exchange is necessary and whether it refreshes or resets existing state. Align the result with the contributing rule that `client/hello` is sent once per connection.

maximmaxim345:
Re-handshake currently repeats both hellos even though connection state survives, contradicting the rule that `client/hello` is sent once per connection. Send `server/activate` directly after re-handshake under the new keys, applying the existing activation rules without resetting connection state. Clarify that both hellos are sent only during initial connection setup and label the pairing diagrams accordingly.

This also resolves the issue for server and client implementations of how to handle changing `client/hello` and `server/hello` messages.

Closes Sendspin/backlog#47

## Open concern: playback during re-pairing

Out of scope: re-pairing an already paired, playing client via `pairing_psk` with unpaired access disabled still requires ending its streams before re-handshake, because playback is not permitted under the pairing PSK in that case. Whether an already trusted connection should retain playback permission while re-pairing remains an open policy question. This predates this change.


chrisuthe (approved):

kahrendt (approved):
Thanks!

PR #288, opened 2026-09-16
Define player buffer-capacity accounting

### LINKED ISSUE #230: Define player buffer-capacity accounting

maximmaxim345:
The player spec leaves servers to interpret which bytes count toward `buffer_capacity` and when they become reusable. Define whole-chunk accounting, including the audio header, through delay-adjusted playback completion, with explicit rules for stream resets, format changes, and output-delay updates. No wire fields or playback transitions change. When decreasing `output_delay_ms` leaves a player holding more than `buffer_capacity`, it must remain operational and should avoid audible interruptions.

Closes #230


kahrendt (approved):
LGTM

PR #289, opened 2026-09-16
Define protocol evolution and unknown-message handling

maximmaxim345:
This defines how future revisions can add features while preserving compatibility, including when core or role versions must change. It also prepares for #126 with role-specific stream completion while keeping existing media behavior.

Clients and servers ignore unknown messages after the initial `server/activate`, while preserving validation and re-handshake restrictions, including messages still arriving with the previous session keys.

Optional informational fields defined by the specification can be sent in either direction without additional negotiation if older receivers can ignore them without changing the existing message's meaning. For example, a future informational field in `client/hello` or `server/hello` could use this exception. New message types, binary IDs, and behavior-changing fields still require confirmed support. This does not permit arbitrary fields or extra bytes that break existing binary layouts.

Role cleanup covers explicit removal, implicit removal, and version replacement. The server must send `stream/end` for active streams before removal, but need not wait for buffered playback to finish. If removal happens in the first activation after re-handshake, send `stream/end` before starting that handshake. Removal clears the affected output, temporary effects, state, and pending state updates.

Potential uses include voice replies, chimes and streamed TTS (#126), new role message types, and companion roles for reporting and controlling native Spotify Connect playback on the device or Bluetooth sources.

Multi-server announcements can be added later without a breaking core change through separately negotiated admission and ownership rules that preserve existing peers' behavior.

<details>
<summary>Multi-server compatibility details</summary>

Existing role negotiation can select a future secondary-announcement role while `client/init.version` and `server/init.version` remain `1`. The client and secondary server need updates. Their new rules must preserve compatibility with an unchanged primary.

The future role must keep each connection's state and cleanup separate and decline unsupported connections before activation rather than taking over playback.

This PR prepares that path but does not implement the role or prove the complete feature. Independent media ducking with unchanged master volume, source-role coexistence, competing announcements, resource use, and audio timing still need a compatibility design.

</details>


messaging.md
```diff
@@ -250,7 +258,11 @@ Servers SHOULD declare the minimal set of activities that reflects the connectio
...
 
-When a `server/activate` removes a stream role (`player`, `artwork`, `visualizer`) that has an active stream from `active_roles`, the server MUST first end that role's output by sending [`stream/end`](#server--client-streamend).
+Before a `server/activate` removes a server-to-client stream role (`player`, `artwork`, `visualizer`, or an application-specific role with such a stream), the server MUST send [`stream/end`](#server--client-streamend) for that role if its stream is active. If the first activation after a [re-handshake](connection.md#re-handshake) will remove such a role, the server MUST send any required `stream/end` before starting the re-handshake.
+
+When applying that activation, the client MUST stop the removed role's remaining output, clear its buffers, and release temporary output effects applied by that role, such as ducking. This applies even if an earlier `stream/end` allowed buffered data to finish playing.
+
+Both requirements apply to explicit removals, implicit removals when the connection is no longer playback-capable, and replacement of an active role version.
```
kahrendt: This is repeating what the next paragraph says pretty closely, could they be combined?

messaging.md
```diff
@@ -27,6 +27,12 @@ All JSON messages have a `type` field identifying the message and a `payload` ob
 
 **Forward compatibility.** Clients and servers MUST ignore unrecognized `payload` fields (keys not defined for the message) rather than treating them as an error. Clients and servers MUST NOT send fields the specification does not define for the message, other than the `_`-prefixed [application-specific role](README.md#application-specific-roles) objects a message explicitly permits, or support objects in `client/hello` for advertised application-specific role versions (e.g., `player@_experimental_support`).
 
+Clients and servers MUST ignore JSON messages with an unrecognized `type`, provided the message is a valid JSON object with a `type` string and a `payload` object. They MUST also ignore binary messages whose ID they do not implement. These rules apply only when the [initial handshake](#communication) and [re-handshake](connection.md#re-handshake) rules allow application messages.
```
kahrendt: The initial handshake link is self-referential to this section. Could we be more direct? It seems like we want something like we want: ignore JSON messages with an unrecognized `type` after the initial server/activate, and not between Noise message 1 and the re-activation.

README.md
```diff
@@ -704,7 +726,11 @@ The server MUST NOT send this message when no targeted streams are active.
 
 ### Server → Client: `stream/end`
 
-Ends the stream for one or more roles. When received, clients MUST stop output and clear buffers for the specified roles. This message is expected to be sent when playback is over and the queue is empty. Specifically:
+Ends the stream for one or more roles. Each side MUST treat the targeted streams as inactive once it sends or receives this message, even if buffered output continues.
+
+For each specified role, clients MUST stop output and clear its buffers unless that role explicitly defines different completion behavior. In that case, clients MUST follow the role's rules, such as finishing playback of buffered data.
+
+For roles following the media queue, this message is expected to be sent when playback is over and the queue is empty. Specifically:
```
kahrendt: "following the media queue" is undefined in the spec (though we do refer to a queue of media in several places)? Maybe switching to "a" instead of "the" makes it refer to a generalized queue versus a specific undefined one. Alternatively, we could define it and specificy what roles follow the media queue.

messaging.md
maximmaxim345: Combined the stream and `server/state` cleanup requirements, with the removal cases listed once.

messaging.md
maximmaxim345: Made the initial `server/activate` boundary explicit. The server still needs to tolerate valid messages encrypted with the previous session keys before Noise message 2, so I kept that exception.

README.md
maximmaxim345: Changed it to “a media queue” in the source. That will cover application specific roles too (like if we want to try out `announce@v1` as a draft role `_wip_announce@v1` before committing it to the spec).

maximmaxim345:
Pushed two commits addressing the review comments and allowing optional informational fields without additional negotiation when older receivers can safely ignore them (such as adding new metadata fields).

I won't be able to use my laptop for the next couple of hours, @kahrendt can you take over this PR in case you find some other issues that need fixing?

kahrendt (approved):
Looks good, thanks!

PR #290, opened 2026-09-17
Clarify an ambiguity in Client Initiated Connections

kepstin:
From the wording of the spec, it's unclear whether the intention was that mDNS was the only method that clients could use to discover a server, or if they were permitted to connect to servers that they had discovered in some other way (e.g. a manually entered `ws://` url).

I think that permitting other methods of discovering the server is fine in some circumstances, so I've added the BCP-14 keyword [SHOULD](https://www.rfc-editor.org/info/rfc2119/#section-3) to the statement to make that clear. 

If the intent was that mDNS is the only method that clients are permitted to use to discover the server, then I'd recommend using the MUST keyword instead.


kepstin:
As an additional note, Javascript running in web browsers is _unable_ to either accept incoming connections or perform mDNS service discovery, so all connections made by JS sendspin clients will be client-initiated connections to a websocket URL provided via an out-of-band mechanism.

